PR #1, opened 2016-05-11
Add dist config for Epic

erikwiffin:
* Moved oAuth logic out of feature/step/oauth and into a dedicated oAuth module.
* Add .pep8 and .pylintrc and associated config files.
* Move bulk of application from root to /testsuite.
* Un-gitignored /static/build so that the docker container no longer needs npm.

See it in action at http://52.39.26.206:9003/?vendor=epic


behave.ini.dist-epic
```diff
@@ -0,0 +1,9 @@
+[behave]
+
+[api]
+url=https://open-ic.epic.com/FHIR/api/FHIR/DSTU2/
+patient=Tbt3KuCY0B5PSrJvCu2j-PlK.aiHsu2xUjUM8bWpetXoB
+
+[auth]
+strategy=none
+client_id=dd24090e-bc30-4a18-8355-fa3664bff207
```
jmandel: Is this just in an inconsistent state? Or can you clarify why the `none` strategy has a `client_id`?

testsuite/oauth/smart.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+ERROR_AUTH_REQUEST = """
+Authenticated POST to {url} returned {status_code}.
+{text}
+"""
+
+
+class SmartStrategy(object):
+    """ Implements the lib.oauth.Strategy interface. """
```
jmandel: Can you clarify whether this strategy is designed to work *only* with SMART's reference server, or whether you intend to generalize it for allscripts, Cerner, Epic etc? If it's to be generalized, the only ways I really know are

##### 1. as a series of form-fill expectations and requests, to be used in a `requests.session`, like:

```python
[
  ["load", authorize_url],
  ["submitForm", ["form.someSelector", {
    "username": "argonaut",
    "password": "argonaut"
  }]
]
```

##### 2. A series of clicks and keystrokes, to be used via webdriver, like:

```python
 [
    {"click": "#txtUsername"},
    {"keys": "ARGONAUT"},
    {"click": "#txtPassword"},
    {"keys": "ARGONAUT"},
    {"click": "#cmdLogin"}
]
```

behave.ini.dist-epic
erikwiffin: Inconsistent state.

testsuite/oauth/smart.py
erikwiffin: I'd like it to work for everyone, but right now I've only tested it against our reference server. What will probably happen is that I'll break out a set of "authorize" strategies as well, and each will be responsible for managing either a headless authorize path (like ours) or some kind of webdriver/session based path.

jmandel:
Sounds good! Happy to merge in this state.

PR #2, opened 2016-05-12
Add authenticated config for Allscripts

erikwiffin:
* Use a refresh token strategy for now to avoid writing a webdriver.
* Also provide un-authenticated config to Cerner.
* Improve display output for failing tests.


behave.ini.dist-allscripts
```diff
@@ -0,0 +1,14 @@
+[behave]
+logging_level=DEBUG
+logging_format=%(message)s
+
+[api]
+url=http://uat-fhir.followmyhealth.com/api/
+patient=8501fd7d-4f46-4eca-a5c0-a5fd0112fe85
+
+[auth]
+strategy=refresh_token
```
jmandel: That's a cute trick :-)

And it buys you a *little bit of time* at least.

testsuite/oauth/refresh_token.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def revoke_access_token(self):
+        """ Request that the oAuth server revoke stored access token.
+
+        Doesn't unset our currently stored data.
+        """
+        post_data = {
+            'token': self.access_token,
+        }
+        response = requests.delete(self._config['token_url'],
```
jmandel: This won't be a _general_ feature of a token endpoint, for what it's worth. Is it supported anywhere?

erikwiffin: No one supports it. Should we just remove this test?

jmandel: Seems wise :-)

PR #3, opened 2016-05-13
Add authenticated Cerner config

erikwiffin:


behave.ini.dist-cerner
```diff
@@ -3,8 +3,12 @@ logging_level=DEBUG
...
 [api]
-url=https://fhir-open.sandboxcernerpowerchart.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/
-patient=1696023
+url=https://fhir.sandboxcernerpowerchart.com/dstu2/d075cf8b-3261-481d-97e5-ba6c48d3b41f/
+patient=1316024
 
 [auth]
-strategy=none
+strategy=refresh_token
+refresh_token=69746025-cd6b-4129-ac7b-b11d04f625b5
```
jmandel: How doe this work if the tokens are only good short-term?

testsuite/templates/launch.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+{% block body %}
+    Loading....
+    <script src="/static/fhir-client.js"></script>
+    <script>
+        FHIR.oauth2.authorize({
+            "client_id": "f662d614-8b45-48c2-92f8-c84a20edccc3",
+            "scope": "launch patient/Patient.read patient/Observation.read patient/Condition.read patient/MedicationOrder.read patient/MedicationStatement.read patient/MedicationDispense.read patient/MedicationAdministration.read patient/AllergyIntolerance.read patient/Procedure.read patient/Immunization.read patient/DocumentReference.read"
```
jmandel: Just FYI, Cerner has indicated that they will eventually support `patient/*.read` -- likely in June. So we won't have to manage this forever.

testsuite/fhir.py
```diff
@@ -0,0 +1,33 @@
...
+    dict :
+        authorize : string
+        token : string
+    """
+    url = '{url}metadata'.format(url=base_url)
+    headers = {
+        'Accept': 'application/json+fhir',
+    }
+    response = requests.get(url, headers=headers)
+    conformance = response.json()
```
jmandel: Can you try using https://github.com/smart-on-fhir/client-py (in pip: `fhirclient==v1.0.3`) to parse results, in general? (If it doesn't work, that would also be important to know.)

testsuite/oauth/refresh_token.py
```diff
@@ -46,13 +41,12 @@ def revoke_access_token(self):
         post_data = {
             'token': self.access_token,
         }
-        response = requests.delete(self._config['token_url'],
+        response = requests.delete(self._urls['token'],
```
jmandel: Are you still planning to nix the `revoke_access_token` method here? (**Eventually** we'll want one but it will pretty much have to be webdriver-controlled to deal with UX variability in the portals.)

behave.ini.dist-cerner
erikwiffin: It doesn't really. I'm thinking that a better solution is actually a "launch/authorize" button on the testing site that walks you through the authorization process and saved a refresh token in a `.refresh_token` file or something like that. That would be another better fix than what we currently have on the road to the webdriver, and most of the logic (like what would power the `/authorized/` endpoint) would actually end up being useful to the webdriver.

testsuite/templates/launch.html
erikwiffin: Should I be making separate /launch pages for the various vendors? What's the best way to put the right client id/scope in here?

testsuite/fhir.py
erikwiffin: Will do.

testsuite/oauth/refresh_token.py
erikwiffin: Still planning on it yeah. It's just hard to delete a code path that might eventually be useful :(

behave.ini.dist-cerner
jmandel: Yes -- though if you're going to put the user through the flow, it might make sense to just save the `access_token` and use *that* as your strategy (since not all servers will support refresh tokens straight off). This is basically what crucible does.

testsuite/templates/launch.html
jmandel: A separate page for vendor is probably fine, since in general we **shouldn't** be using this ("EHR Launch") approach to launch anyway. We want to be using the "Standalone Launch". Right now, Cerner only supports EHR Launch, so we need this for Cerner. Maybe there will be a few others too.

Have you thought about how to manage these launch scenarios in cases where >1 user might be running the test suite at the same time?

behave.ini.dist-cerner
erikwiffin: I think all the servers support refresh tokens right now. Cerner's are "online_access" tokens, but still follow the refresh token path.

testsuite/templates/launch.html
erikwiffin: That's a good point. Makes me think that the eventual token shouldn't be stored in a file somewhere, but in the user session (or maybe fall-back to file if there is no session, so the cli script still works).

jmandel: User session seems like a good default place for the config data to live, yeah.

(BTW, if I can make a little bit of time to geek out, I'm thinking about writing a wrapper around https://github.com/behave/behave/blob/master/behave/formatter/json.py to pass individual results as JSON snippets through a websocket connection, rather than streaming a single giant JSON document to stdout.)

erikwiffin: That would be sweet! It doesn't look like flask natively supports websockets, but https://github.com/miguelgrinberg/Flask-SocketIO exists and looks pretty easy to use.

behave.ini.dist-allscripts
```diff
@@ -10,5 +10,6 @@ patient=8501fd7d-4f46-4eca-a5c0-a5fd0112fe85
 strategy=refresh_token
 client_id=4e3ecd32-b17c-47ed-bbb0-ff60e870be1a
 client_secret=QYuz2cPx6e9Th/UkYrTwuBjChAYa8aebkT9VVsgC211NboKAqLt0BZwrxSp+De/oS4iSzUi2nePZI6NAF+UoIw==
-refresh_token=eyJhbGciOiJIUzI1NiIsImtpZCI6InRlbXAtZGVmYXVsdCIsInR5cCI6IkpXVCJ9.eyJuYmYiOjE0NjMwNjc4MDgsImV4cCI6MTQ5NDYwMzgwOCwiaWF0IjoxNDYzMDY3ODA4LCJ1bmlxdWVfbmFtZSI6IjRlM2VjZDMyLWIxN2MtNDdlZC1iYmIwLWZmNjBlODcwYmUxYSIsIlVzZSI6IlJlZnJlc2giLCJVc2VyIjoiODUwMWZkN2QtNGY0Ni00ZWNhLWE1YzAtYTVmZDAxMTJmZTg1Iiwic2NvcGUiOiJsYXVuY2gvcGF0aWVudCBwYXRpZW50LyoucmVhZCIsImlzcyI6Imh0dHBzOi8vbXVsb2dpbi5mb2xsb3dteWhlYWx0aC5jb20iLCJhdWQiOiJodHRwczovL211bG9naW4uZm9sbG93bXloZWFsdGguY29tIn0.OF-a--gMzZgqDugDE_TssFPIPxpmmI1Z6PohWfFBpHI
-url=https://fmhmulogin.azurewebsites.net/api/
+scope=launch/patient patient/*.read offline_access
+redirect_uri=https://tests.demo.syncfor.science/authorized/
```
jmandel: At this point, given we're passing these via POST and parsing fields out of them ourselves, these files might as well be JSON. (Am I missing something?)

testsuite/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+@app.route("/")
+def index():
+    return render_template('index.html')
+
+
+@app.route('/tests.json', methods=['POST'])
+def tests():
+    from behave.configuration import Configuration
```
jmandel: I'm not familiar with this style of importing from within a function — is there a reason it's desirable? http://legacy.python.org/dev/peps/pep-0008/#imports and http://stackoverflow.com/questions/1024049/is-it-pythonic-to-import-inside-functions seem down on it, but I'm not well-versed on this topic.

testsuite/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+    uris = get_oauth_uris(config['api']['url'])
+    strategy = RefreshTokenStrategy(
+        client_id=config['auth']['client_id'],
+        client_secret=config['auth']['client_secret'],
+        redirect_uri=config['auth']['redirect_uri'],
+        urls=uris,
+        refresh_token=config['auth']['refresh_token'],
+        basic=config['auth'].get('basic', False)
+    )
+    strategy.upgrade_authorization_code(request.args.get('code'))
```
jmandel: Just FYI https://tools.ietf.org/html/rfc6749 uses the language "exchange an authorization grant for an access token".

testsuite/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+
+@app.route('/launch/epic/', methods=['POST', 'GET'])
+def launch_epic():
+    from testsuite import config_reader
+
+    config = config_reader.get_config('epic')
+    return render_template('launch.html', config=config)
+
+
+@app.route('/launch/cerner/', methods=['POST', 'GET'])
```
jmandel: Given the two examples we have so far, looks like you could DRY this out to take a path parameter (`cerner` or `epic`).

testsuite/config_reader.py
```diff
@@ -1,13 +1,16 @@
...
+from flask import request, session
 
 
-def get_config():
+def get_config(default='smart'):
     config = configparser.ConfigParser()
 
     try:
-        config.read_string(request.form['config'])
+        vendor = request.form.get('vendor', default).lower()
```
jmandel: Grabbing state from `request` here is kind of evil (took me a while to figure our what was happening when I was working on the websocket integratoin).

testsuite/__init__.py
erikwiffin: Thanks! If there's language in the spec I definitely want to be using it.

testsuite/config_reader.py
erikwiffin: Yeah, it's really unfortunate, but the way behave actually invokes the test suite makes it really difficult to be explicit about dependencies like that. If I remember right, they're basically calling `eval` on environment.py. It makes sense on one hand, since it makes it impossible to accidentally share state between tests, but it would be really useful if I could pass environments into the test suite :/

jmandel: My technique was just to pass details in the `configuration.defaults` dict
(instantiated in app.py), and read from that in config_reader. That seemed
at least somewhat less implicit/surprising (though still somewhat
implicit/surprising).

PR #4, opened 2016-05-18
Code cleanup + unit tests

erikwiffin:
* Moves vendor configuration from behave.ini.dist-vendor to config/vendor.yml. Behave configuration is no longer tied to vendors, since that doesn't really make any sense.
* Behave context is provided with a config dict and doesn't do attempt to parse it out at all. This makes it easier to share configuration between behave and non-behave contexts.
* Wrote a couple *really* basic tests with py.test.


jmandel:
Nice! Can you add documentation about how to invoke these new test suite tests? (My this feels meta.)

erikwiffin:
We need to go deeper!

PR #9, opened 2016-05-27
Add config for Athenahealth

erikwiffin:


jmandel:
Great. Would clarify the security note to indicate what's happening: these clients are registered only to access testing environments in which no protected health information is present.

PR #17, opened 2016-06-15
Include plan in snapshot

jmandel:


features/environment.py
```diff
@@ -24,9 +24,20 @@
 
 
 def before_all(context):
-    config = get_config(context.config.vendor)
 
-    context.config = config
+    plan = []
+    for feature in context._runner.features:
```
erikwiffin: Since right now we only need the count of features, I'd be more inclined to leave out this whole loop.

features/environment.py
```diff
@@ -24,9 +24,20 @@
...
+        scenariolist = []
+        plan.append({'name': str(feature), 'scenarios': scenariolist})
+        for scenario in feature.scenarios:
+            steplist = []
+            scenariolist.append({'name': str(scenario), 'steps': steplist})
+            for step in scenario.steps:
+                steplist.append({'name': str(step)})
+
+    context.config.plan = plan
+    context.config = get_config(context.config.vendor)
```
erikwiffin: Clearly this does work, but I feel like it shouldn't. Wouldn't setting `context.config` here blow out the `context.config.plan` property?

testsuite/formatter.py
```diff
@@ -8,10 +8,14 @@ class ChunkedJsonFormatter(PrettyJSONFormatter):
     """
     def __init__(self, a, b):
         super(ChunkedJsonFormatter, self).__init__(a, b)
-        self.snapshot = []
+        self.snapshot = {
```
erikwiffin: Instead of converting snapshot into a dictionary, why not make `plan` a top level property as well? Then pass them both independently through `on_snapshot` and `emit`. Mostly I'm thinking that `status` already has a meaning in the behave context, and I don't want them to get confused.

erikwiffin:
Did you want me to apply this to my websocket pull request? Or were you going to merge that and then have me merge this?

features/environment.py
jmandel: I was assuming we wanted to create the whole scaffold for response, rather than just the count, so we'd have a grid with cells pre allocated.

testsuite/formatter.py
jmandel: That sounds like a good idea! Mostly I just wanted to work out the data  flow with this PR, to figure out where to grab what we need.

features/environment.py
jmandel: I was assuming we wanted to create the whole scaffold for response, rather than just the count, so we'd have a grid with cells pre allocated.

PR #19, opened 2016-06-17
Feature/conformance statement

erikwiffin:


features/steps/s4s.py
```diff
@@ -28,6 +37,61 @@ def parse_mu_ccds_mapping(mu_ccds):
 register_type(MU_CCDS=parse_mu_ccds_mapping)
 
 
+@given('I have a valid conformance statement')
+def step_impl(context):
+    assert context.conformance is not None, \
```
jmandel: We had talked about defaulting to an "everything is supported" conformance statement in the event that a server was missing its conformance statement altogether, or in the event that it listed no resources of any kind. Is this a non-issue (because vendors actually already have working conformance statements)? Or: how many vendors just stop working altogether if you apply this hard-line logic?

erikwiffin: This statement only gets run in `10-sync-for-science.feature`. All the actual resource tests use `this server supports {resource_type}` which has the early escape of:

```
# If there is no conformance statement, assume this resource is supported
if context.conformance is None:
    return
```

jmandel: Got it. I keep forgetting that an empty return is a passing test :-)

PR #21, opened 2016-06-20
Summarize results + keyboard nav

jmandel:
image


app/error-navigation.js
```diff
@@ -0,0 +1,48 @@
...
+function moveError(by){
+  var i = errorNavigation.index;
+  var len = errorNavigation.errors.length;
+  var next = (((i + by) % len) + len) % len; // ensure positive, bounded
+
+  errorNavigation.index = next;
+  scrollToError();
+}
+
+document.addEventListener("keydown", function(e){
```
erikwiffin: As much as possible, I like to enforce side-effect-free imports, so this event listener would be better in main.js or in some kind of `init()` method that gets exported.

app/error-navigation.js
```diff
@@ -0,0 +1,48 @@
...
+
+function scrollToError(){
+    $("#"+dashify(errorNavigation.errors[errorNavigation.index].location))[0]
+    .scrollIntoView();
+}
+
+function moveError(by){
+  var i = errorNavigation.index;
+  var len = errorNavigation.errors.length;
+  var next = (((i + by) % len) + len) % len; // ensure positive, bounded
```
erikwiffin: And I just learned something new and disappointing about javascript.

jmandel: ;-) Yeah, me too (yesterday). And agreed, `init` would be much cleaner!

app/error-navigation.js
jmandel: Agreed, `init` would be much cleaner!

PR #22, opened 2016-06-21
First pass at celery-ifying test runner

erikwiffin:
Probably not ready to merge, but here's a first pass at getting celery to work. It seems to run all the way through without issue.

* I figured out how to use supervisord with a "first run" process. A clever combination of `autostart=false` and `autorestart=false`.
* will want to move to using actual rooms instead of the `request.sid`, since that changes in the event of a disconnection.
* I left a todo for this, but there's a problem automatically serializing the test snapshot, but if I json/un-json it, it works.
* config management is a little out of control, maybe that gets fixed here, maybe I try to wrangle that in it's own commit.


jmandel:
The message handlers are now *very clean* — nice!

Agreed, we'll need to work out the environment/config management approach.

PR #24, opened 2016-06-22
More efficient docker build

jmandel:


.gitignore
```diff
@@ -6,6 +6,7 @@ build
 *.egg-info
 ghostdriver.log
 dump.rdb
+venv
```
erikwiffin: I've been using `ENV` instead of `venv` for the python virtual environment.

Dockerfile
```diff
@@ -21,11 +21,14 @@ RUN wget --quiet https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1
 RUN tar xvfj /opt/phantomjs-2.1.1-linux-x86_64.tar.bz2 && \
     ln -s /opt/phantomjs-2.1.1-linux-x86_64/bin/phantomjs /usr/bin/phantomjs
 
-# Copy the codebase
-COPY . /demo-test-suite
+RUN mkdir /demo-test-suite
```
erikwiffin: What's the point of running mkdir instead of just starting by copying?

jmandel: I was trying to avoid creating a *file* called /demo-test-suite whose
contests look like the lines of requirements.txt; have I misunderstood the
semantics of "COPY"?

erikwiffin: Maybe I'm getting the "wrong" result because of my local volume mounting, but I'd expect `COPY . /demo-test-suite` to copy the contents of the current directory into a /demo-test-suite directory. Shouldn't have anything to do with requirements.txt either way though.

PR #25, opened 2016-06-22
Move phantomjs download URL from bitbucket to github

erikwiffin:


erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/35. 
According to https://github.com/ariya/phantomjs/issues/13953, bitbucket rate limits downloads, and the phantomjs folks are still looking for a better place to host downloads.

PR #26, opened 2016-06-22
Feature/fhir python validation

jmandel:


jmandel:
There are  definitely opportunities to "DRY" up the individual feature files here. Making an addition currently requires across-the-board changes.

PR #28, opened 2016-06-23
Feature/oauth tests

erikwiffin:


supervisord.conf
```diff
@@ -39,8 +39,16 @@ command=/usr/bin/env celery -A testsuite.tasks worker --config=testsuite.celeryc
 directory=/usr/src/app
 environment=C_FORCE_ROOT="yes"
 autostart=false
+stdout_logfile=/dev/stdout
```
erikwiffin: Fixes https://github.com/sync-for-science/tracking/issues/37.

testsuite/fhir.py
```diff
@@ -27,22 +27,20 @@ def get_conformance_statement(base_url):
         raise
 
 
-def get_oauth_uris(base_url):
+def get_oauth_uris(conformance):
```
erikwiffin: Fixes https://github.com/sync-for-science/tracking/issues/36.

Dockerfile
```diff
@@ -31,6 +31,7 @@ COPY . /usr/src/app
 
 RUN git clone https://github.com/smart-on-fhir/client-py validator && \
     cd validator && \
-    git reset --hard fd67e6a8717edfc63fb7293405a1c45a4cfac8c2
+    git reset --hard fd67e6a8717edfc63fb7293405a1c45a4cfac8c2 && \
+    touch __init__.py
```
erikwiffin: Pylint throws errors when it tries to import from directories without `__init__.py` files.

jmandel: Thanks! Does pylint run as part of the travis tests BTW? I don't remember this generating an error (but I may have just been oblivious).

app/summarize.js
```diff
@@ -1,26 +1,40 @@
+var _ = require('underscore');
+
 module.exports = function(event) {
   var errors = [];
+  // Deep-clone the event plan
   var summary = JSON.parse(JSON.stringify(event.plan));
+
   summary.forEach(function(f, i){
+    // Plan and snapshot offsets may not match, get the correct key
+    var fkey = _.findKey(event.snapshot, function (feature) {
```
jmandel: Can you clarify when these don't match? I've definitely misunderstood something here.

app/templates/features.hbs
```diff
@@ -9,6 +9,22 @@
                 <div class="steps">
                 {{#each steps}}
                     <span class="{{ result.status }}">{{ keyword }} {{ name }} {{result result.status}}</span>
+                    {{#if table}}
```
jmandel: When/where is `table` ever set here?

Dockerfile
erikwiffin: It hasn't been, I have it set as a write hook in vim. I'll see how hard it is to add it to travis though.

app/summarize.js
erikwiffin: I've added ```
 tags=smart
 show_skipped=no
``` to my local behave.ini so that I can test *only* the oauth part of the process, makes things run faster. In this case, the snapshot only contains the tests I've run.

app/templates/features.hbs
erikwiffin: image

jmandel: Ohhh, the "table" property is generate python-side! Thanks.

PR #29, opened 2016-06-25
Generalize OAuth approval to clicking+typing

jmandel:


testsuite/oauth/authorize/smart.py
```diff
@@ -11,5 +11,4 @@ class SmartAuthorizer(base.AbstractAuthorizer):
         authorize_url (string): The vendor's authorize endpoint.
     """
     def __init__(self, config, authorize_url):
-        self.config = config
-        self.authorize_url = authorize_url
+        super().__init__(config, authorize_url)
```
erikwiffin: Why does Smart still need an authorizer?

jmandel: It doesn't. this file should have been deleted :-)

testsuite/oauth/authorize/base.py
```diff
@@ -128,9 +158,10 @@ def open(self):
         self.browser = self._browser()
 
     def close(self):
-        """ Close the selenium webdriver.
+        """ Close the virtual display.
```
erikwiffin: Selenium webdriver is still being closed here. Maybe a more generic comment like "close the virtual browser"?

testsuite/oauth/authorize/base.py
```diff
@@ -143,3 +174,5 @@ def find(self, selector):
         """ Shorthand for jQuery style element selecting.
         """
         return self.browser.find_element_by_css_selector(selector)
+
+class Authorizer(AbstractAuthorizer): pass
```
erikwiffin: It looks like you could just get rid of the Abstract Authorizer, move this module up a level, and have an Authorizer class that everything uses. NoneAuthorizer can inherit a concrete class just as well.

testsuite/oauth/authorize/base.py
```diff
@@ -1,10 +1,13 @@
 """ Authorize the SMART API.
 """
+import time
 from abc import ABCMeta
 from urllib import parse
+from pyvirtualdisplay import Display
```
erikwiffin: Nitpick, but: I generally try to alphabetize the import statements (so `time` would go last) and they are also organized into groups: stdlib, pip modules, and local modules. So pyvirtualdisplay would go down with selenium.

https://www.python.org/dev/peps/pep-0008/#imports

testsuite/oauth/authorize/base.py
jmandel: Agreed!

testsuite/oauth/authorize/base.py
jmandel: Fair enough -- I will follow all of your nitpicks if you update the lints to fail me for breaking them :-)

PR #30, opened 2016-06-27
Add test for too-long state parameter

erikwiffin:


config/allscripts.yml
```diff
@@ -11,11 +11,12 @@ auth:
...
   steps:
-    - find: "[translate='DEFAULT_LOGIN_BUTTON']"
-    - click
-    - find: "#UserName"
-    - type: "s4s_6-22-16"
-    - find: "#Password"
-    - type: "s4s!2345"
-    - find: "[translate='Login_LogIn']"
-    - click
+    - element: "[translate='DEFAULT_LOGIN_BUTTON']"
```
erikwiffin: Hopefully this isn't too much bikeshedding, but I hope this is a more intuitive step configuration format. Each step selects an element, and then performs an action on it. The action is a "key" in the yaml dictionary, so it can take any parameters it needs. Currently defined actions are `send_keys` and `click`. Steps can also be marked as optional, and if the element can't be found, they'll be silently skipped.

features/20-standalone-launch.feature
```diff
@@ -31,6 +31,14 @@ Feature: Follows the SMART standalone launch sequence
         When I ask for authorization without the state field
         Then the response code should not be 200
 
+    Scenario: EHR evaluates authorization request - very long "state"
```
erikwiffin: I don't know if this is actually something we want to test for, but I wanted to put together the scaffolding for testing the full authorization process, and this seemed like an easy thing to throw in there.

testsuite/oauth/authorize.py
```diff
@@ -0,0 +1,236 @@
...
+
+        try:
+            self._validate_state(response)
+            self._validate_code(response)
+        except AssertionError as err:
+            raise ValidationErrorException(str(err), self.browser)
+
+        return response['code'][0]
+
+    def ask_for_authorization(self, parameters):
```
erikwiffin: Since this isn't a "skeleton" anymore, I renamed these methods to match the language on http://docs.smarthealthit.org/authorization/ a bit better.

testsuite/oauth/authorize.py
```diff
@@ -0,0 +1,236 @@
...
+        return self.substring in driver.current_url
+
+
+class AuthorizationException(Exception):
+    """ An Error occured during the authorization process.
+    """
+    def __init__(self, message, browser):
+        # Save a screenshot so we can see what's up
+        path = 'testsuite/static/screenshots/{0}.png'.format(uuid.uuid4())
+        browser.save_screenshot(path)
```
erikwiffin: I'm pretty proud of this, hope the vendors find it useful. http://tests.dev.syncfor.science:9003/static/screenshots/7b23c3fa-1c89-497b-a69c-74704a51b20a.png

config/allscripts.yml
jmandel: Haha, I had a format **exactly** like this, except I used the word `find` instead of `element`. I found myself missing/misreading the YAML arrays/dicts, and then I realized I could flatten the whole structure out and make what I thought was a more explicit sequence. The empty `click:` kind of weirds me out if you want to bikeshed (so much so that in my version I said `click: left` just to avoid it).

But I'm happy enough with this :-) The main goals was just steps-as-data instead of steps-as-code.

features/20-standalone-launch.feature
jmandel: I wouldn't think it's a high-yield test; does anybody fail it? Doesn't hurt anyway, as a scaffolding!

testsuite/oauth/authorize.py
```diff
@@ -0,0 +1,236 @@
...
+
+        Step 2 of the SMART authorization process. Usually this would include
+        logging in and clicking an "authorize" button.
+        """
+        for step in self.config.get('steps', []):
+            self._execute_step(step)
+
+        # Some vendors implement an AJAX based login procedure.
+        wait = WebDriverWait(self.browser, AUTHORIZE_TIMEOUT)
+        wait.until(CurrentUrlContains(self.config['redirect_uri']))
```
jmandel: Lovely!

testsuite/oauth/authorize.py
jmandel: Darn, this is a nice touch!! I love it.

jmandel: Note that in current status, I can get errors on the server side like:

```
    self._execute_step(step)
  File "/usr/src/app/testsuite/oauth/authorize.py", line 161, in _execute_step
    raise ElementNotFoundException('Element is hidden', self.browser)
testsuite.oauth.authorize.ElementNotFoundException: ('Element is hidden', '/static/screenshots/777e8c08-1844-42da-bf27-49c12c3a95f3.png')
```

but unless they occur in a "When I ask for authorization with the following override" scenario, they don't bubble up to the UI (and the UI spins forever). Would be good to emit a "failed" message" when any uncaught exception occurs, so the UI can render it.

features/20-standalone-launch.feature
erikwiffin: The screenshot from further down is actually Allscripts failing this test. I don't know if it's the length or the `'` in the state string, but it does actually return results.

testsuite/oauth/authorize.py
erikwiffin: Good catch. I'll look into that. It's probably an exception getting thrown in the `before_all` hook, and it's a bit tricky to get those to bubble up in a way that make sense. For the most part I've been trying to catch them and proceed as if oauth isn't enabled.

config/allscripts.yml
erikwiffin: I get what you mean about the empty `click:`. I was tempted to fill them with `click: True` or something like that, just as a placeholder. On the other hand, interchanging dictionaries with a single key (`find:`) and scalars (`click`) was terrifying :D

jmandel:
Thanks -- just tried these out and this flows through perfectly!

PR #37, opened 2016-07-01
Use redis to cache retrieved FHIR resources

erikwiffin:


jmandel:
Wow. I was expecting, like, a python dict ;-)

erikwiffin:
I tried a python dict at first, and I think the resetting of state between tests was causing issues. We already had redis running though, so hooking that up was pretty easy.

jmandel:
Wait, just so I understand what happened with the dict: do you mean
resetting the cache between features of a single run was causing trouble?
(Why would the cache be reset between features of a single run?)

jmandel:
BTW, this is a major speed improvement on some of the servers with large
data volume!

erikwiffin:
The way that behave determines feature/scenario scope was reseting the cache dict each run. It basically made a copy of the cache, gave it a step, the step filled the copy, and then the copy was thrown away.

PR #38, opened 2016-07-04
Add UI for toggling behave tags

erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/42.


runserver.sh
```diff
@@ -1,4 +1,3 @@
 #!/bin/sh
 
-pip install -r requirements.txt
```
erikwiffin: This was a development hack to make sure all my pip packages were installed, shouldn't have been pushed in the first place.

PR #40, opened 2016-07-05
Support data flow under positive + negative approval scenarios 

Cphrampus:
From the previous discussion, we wanted to have a go at laying out the wording of the features for testing. This is just a first try, so if there is anything that should added or modified to improve these, any feedback would be greatly appreciated.


erikwiffin:
This looks like a good first step. One thing that'll be necessary is outlining a few more of the preconditions required for the tests to work, as well as being a little bit more explicit about requesting resources.

```gherkin
Scenario: Authorized to view Allergies and intolerances
    Given I OAuth is enabled
    And I am not logged in
    When I authorize the app allowing access to
        | type |
        | allergies-and-intolerances |
    And I request a Allergies and intolerances response
    Then the response code should be 200

Scenario: Not authorized to view Allergies and intolerances
    Given I OAuth is enabled
    And I am not logged in
    When I authorize the app denying access to
        | type |
        | allergies-and-intolerances |
    And I request a Allergies and intolerances response
    Then the response code should be 403
```

This would also likely require an update to the `steps` list in the [allscripts config](https://github.com/sync-for-science/test-suite/blob/master/config/allscripts.yml) file that includes a concept of "tags" that can enable or disable actions during the authorization process:

```yaml
steps:
  - element: "#UserName"
    send_keys: "s4s_6-22-16"
  - element: "#Password"
    send_keys: "s4s!2345"
  - element: "[translate='Login_LogIn']"
    click: True
  - element: "input.meds"
    click: True
    tag: Approve.medications
  - element: "input.allergies"
    click: True
    tag: Approve.allergies-and-intolerances
  - element: "input.pt"
    click: True
    tag: Deny.patient
```

Cphrampus:
Ok, I had left those out because the request was in the before_feature code, but it makes sense to have to explicitly request them if we are logging out to allow/deny a certain tag.

I have one question. Is there a particular benefit to using the type array in your example as opposed to just a variable verified to be in the CCDS_TAGS list, like it is used in s4s.py? Especially if we are only testing a single authorization.

jmblakl:
One thing that I'm wondering about here is repeat-ability of these tests.  Right now, if a patient user grants an app access to her/his resources the grant screen will only show the first time that user logs in (or at least until the user disallows the access) . This is actually the same way major OAuth servers operate like facebook or google. One way to get around this (short of requiring a log in to disallow manually) is to create a new app registration.  I wonder If we should have a supplementary step which executes before most anything else:

Scenario: Authorized to view Allergies and intolerances

>Given I OAuth is enabled
>And I have created a new client app
>And I am not authorized
>When I authorize the app allowing....etc

erikwiffin:
> Is there a particular benefit to using the type array in your example as opposed to just a variable verified to be in the CCDS_TAGS list, like it is used in s4s.py?

Using the table structure makes it easier to add multiple approvals/denials, but if it doesn't look like you'll be testing anything like that, using a variable is probably good enough.

> Right now, if a patient user grants an app access to her/his resources the grant screen will only show the first time that user logs in

Is that true even if the app directs the user to the authorize endpoint? If so, creating a new client app would probably remove the need for a `And I am not authorized` step. Do you think it makes more sense to create a new client app each time or to hit some kind of "de-authorize" endpoint to reset the permissions?

jmandel:
We're also expecting that a "de authorization" UI will be built into each
portal, even if there's no API for it.

So alternatively we could have a "deauthorize" step that implements clicks
through that process. Otherwise, we'd be requiring every vendor to
implement a "reset" or "client registration" api (I know Allscripts has
this working today, but most vendors don't/won't, and we're not viewing it
as a requirement which makes it hard for us to rely on in testing).

jmblakl:
We don't have a de-authorize endpoint today, but we are already adding it for the patient to utilize anyway so that may be our best approach.  We could do either.  One of the core requirements of the MU3 EHR api access is to "not have app registration a barrier to data access" [paraphrase] so we were figuring that all vendors needed some sort of client registration endpoint.  

The de-authorize endpoint makes sense.

jmandel:
Indeed -- registration can't be a barrier, but it could well involve asking
an app developer to fill out a web form (nothing in the current regulations
requires an API yet).

So my sense is that the most reliably testable approach would probably be
to leverage the patient-facing deauthorization process.
On Jul 6, 2016 14:19, "hello" <notifications@github.com> wrote:

We don't have a de-authorize endpoint today, but we are already adding it
for the patient to utilize anyway so that may be our best approach. We
could do either. One of the core requirements of the MU3 EHR api access is
to "not have app registration a barrier to data access" [paraphrase] so we
were figuring that all vendors needed some sort of client registration
endpoint.

The de-authorize endpoint makes sense.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/sync-for-science/test-suite/pull/40#issuecomment-230859601>,
or mute the thread
<https://github.com/notifications/unsubscribe/AATHAey25l1pPwt714ZeE20SRDToN3VKks5qS_GxgaJpZM4JFhWm>
.

jmblakl:
Hey, guys.  I have been super busy the past week. Who is working on what? Who's court is the ball in?  I had to take my eye off this for a while...sorry.

jmandel:
Thanks @jmblakl! (BTW did you know that [your Github name is "Hello"](https://github.com/jmblakl
)?)

I think we're at a point where we've proposed some tweaks to the way the authorization steps are written in `.feature`, and the way these steps are backed by webdriver commands in a config file.

@Cphrampus, would you be willing to update this PR to use this step syntax and also to show what a config file would look like for clicking through to approve/deny specific scopes in the FMH testing screen? Then we could guide you through the process of building support for making these actually run (and we might be able to help fill some of the gaps to keep you productive).

jmblakl:
I was not aware.  Thanks!

@Cphrampus, let me know if you need help prioritizing this with the other things we assigned to you.  As our intern this SHOULD NOT interfere with getting us our coffee. :)

But seriously, we can talk about priority if you need.

Cphrampus:
I've pushed the updates that I made to the files, as per the previous discussion. I only have the single Deny.Patient, as it seemed like would be a single call to clear all previous approvals, but I am not sure if this was the intention. I am not really sure where the conversation ended in regards to the deauthorization process, so that is not currently in the feature files. Please let me know if that is something that needs to be added.

config/allscripts.yml
```diff
@@ -10,13 +10,53 @@ auth:
...
+      tag: Approve.procedures
+    - element: "input.immunizations"
+      click: True
+      tag: Approve.immunizations
+    - element: "input.documents"
+      click: True
+      tag: Approve.patient-documents
+    - element: "input.pt"
+      click: True
+      tag: Deny.patient
```
jmandel: Just to check my understanding: does this indicted that the approval screen in FMH will have all resources unchecked to start, except Patient which will be checked? So taking no action approves Patient only ml

jmandel: Also, I'd expect to see a final "approve" click, with no tags as signed (meaning it always runs). Does this make sense?

Cphrampus:
The final approve click does make sense, I had forgotten to add that, I added that to the end as just:
```
- element: "input.Approve"
click: True
```
which, if my understanding of what you said is correct, should run regardless.


Yes, the intention is that everything is unchecked by default, then we can go through one type at a time, request an item of that type, make sure that we do not get anything back(except maybe an unauthorized "403" response code), approve that, and check whether we can then get an item of that type.

jmblakl:
Cody, I don't think I understood your question today.   All boxes will be
checked by default

On Tue, Jul 12, 2016, 5:13 PM Cody Phrampus <notifications@github.com>
wrote:

> The final approve click does make sense, I had forgotten to add that, I
> added that to the end as just:
>
> - element: "input.Approve"
> click: True
>
> which, if my understanding of what you said is correct, should run
> regardless.
>
> Yes, the intention is that everything is unchecked by default, then we can
> go through one type at a time, request an item of that type, make sure that
> we do not get anything back(except maybe an unauthorized "403" response
> code), approve that, and check whether we can then get an item of that type.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/sync-for-science/test-suite/pull/40#issuecomment-232198212>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ADr4gCvO8YBqwaayt1ZstvT-P-ix-fNtks5qVBGTgaJpZM4JFhWm>
> .
>
-- 
Justin Blakley

jmandel:
Okay, in that case the only actions you should need to find are they deny actions. You won't need to define any specific approve steps in your list.

jmblakl:
@jmandel - do we have a story on how scopes are going to work for our scenario -- where vitals, labs, and social hx are all considered observations, but we consider them as different things?  How will these tests work if the scope only describes observations and not the subcategory.

Cphrampus:
I have updated the steps to instead deny each individually. However, I still have a "Deny.patient" tag that I had copied from the example given, we are not sure if that is necessary or what it would be for, we have a separate demographics tag, which is what we thought the patient tag might be. Is this something different that we need for something, or was that just something in the example that didn't actually map to a specific authorization?

jmandel:
@jmblakl the story here is that the app asks for `patient/*.read` (i.e. to see all the data that the patient can access in the portal). Internally, the portal may have various ways to allow a patient to limit, hold back, filter, or redact data. We don't try to model those workflows or decisions in S4S; the app gets a token and will try to fetch each of the data categories in turn. (In terms of requirements, we **do** want to give the app some indication that **some data** are being withheld — e.g. an additional scope like `redacted` — without trying to say precisely which data). So as far as the wire protocol goes, that's the whole story today.

Now, here on this PR we're talking about **testing** . Specifically, we're talking about augmenting the test suite to try out certain redaction flows inside of the portal UI. Not all portals will support exactly the same redaction/filtering workflows, and that's fine. But the goal here is to define some rough approach to testing that, and we're aligning these specifically with the data categories in the MU3 Common Clinical data set. So, with you and Cody, we're outlining a set of tagged "approve" and "deny" steps that correspond 1:1 with these categories. That why we've proposed tags like "Approve.vital-signs" and "Approve.lab-results" rather than a generic "Approve.observations". Does this make sense?

jmblakl:
Yes, thanks for the clarification

PR #41, opened 2016-07-05
Allow vendors to provide a config override

erikwiffin:


jmandel:
This works great! Two quick notes;

 * size of the "override" box is not wide enough (sample payload doesn't fit). I'm not sure how to fix this. I mean, a modal pop-out would be less functional for power users, because you couldn't just paste-and-go as easily. Maybe the simplest thing is to move the example into a "tempalte.yml" file (or some such) in our github rep, and just link to the example (target=_blank) from the "More options" screen? Too complicated?



 *  The "More options" link should change somehow when open/closed. Maybe "Hide more options"? Or make it into a button that takes on an enabled state when the options are open.

 * In UI, might say "Custom configuration" instead of "override"? What do you think?

erikwiffin:
Fixes: https://github.com/sync-for-science/tracking/issues/41

PR #43, opened 2016-07-06
Log FHIR resource requests

erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/28.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,4 +29,6 @@ COPY requirements.txt /usr/src/app/
 RUN  pip install --no-cache-dir -r requirements.txt
 COPY . /usr/src/app
 
+ENV ES_URL "https://search-s4s-logs-xsjsafiwd7vkpiucmjqmdjkp7y.us-west-2.es.amazonaws.com/test-suite/log/"
+
 CMD supervisord -c supervisord.conf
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -29,4 +29,6 @@ COPY requirements.txt /usr/src/app/
 RUN  pip install --no-cache-dir -r requirements.txt
 COPY . /usr/src/app
 
+ENV ES_URL "https://search-s4s-logs-xsjsafiwd7vkpiucmjqmdjkp7y.us-west-2.es.amazonaws.com/test-suite/log/"
```
jmandel: Can we leave this as a [deploy-time config setting](https://github.com/sync-for-science/reference-stack-docker/blob/master/deploys/demo.syncfor.science/docker-compose.yml#L48)? That will mean that our "prod" db only sees results from prod runs.

features/environment.py
```diff
@@ -28,6 +28,7 @@
     'immunizations',
     'patient-documents',
 }
+DEFAULT_ES_URL = 'https://search-s4s-logs-xsjsafiwd7vkpiucmjqmdjkp7y.us-west-2.es.amazonaws.com/test-suite/log/'
```
jmandel: Same here, I think -- default should be no logging (especially since the test suite could be used in-house by vendors)

PR #45, opened 2016-07-13
Improve stability by dropping chromedriver

erikwiffin:


testsuite/celeryconfig.py
```diff
@@ -9,3 +9,12 @@
 # A whitelist of content-types/serializers to allow.
 CELERY_ACCEPT_CONTENT = ['json']
 CELERY_TASK_SERIALIZER = 'json'
+
+# Limit celery tasks so that if one hangs it doesn't lock the server forever
```
jmandel: Will you add the prefetch limit here too?

jmandel: Will you add the prefetch limit here too?

PR #46, opened 2016-07-19
Save/load state in hash

jmandel:
Just to make it easier to create links that include all the relevant state for the test runner.


erikwiffin:
It feels like a bad idea to just stick a big JSON blob in a URL hash, but I can't think of any actual reasons not to.

* Seems to handle quotes and other weird characters just fine.
* Pasting the sample yaml config brings to total url length up to 833, which is still pretty far from the max url length.
* It doesn't expose anything that isn't exposed anyways through the UI.

PR #49, opened 2016-07-20
Fix/state hash decode

jmandel:


erikwiffin:
What about base64 encoding the JSON string so that there's never any weird transmission errors?

jmandel:
That's true, but I do like the ability to read the parameters. If we wind
up needing to go back and fix this again, I will give up on readability :-)

PR #50, opened 2016-07-20
Add support for tagged steps

erikwiffin:
I've added a `.feature` for immunizations to my beta. We don't yet have support for optional tests so that'll probably need to happen before the Allscripts stuff gets merged.

[custom config that passes "allow" but fails "deny"](http://goo.gl/Is13iO)

### DIFF
--- a/features/steps/oauth.py
+++ b/features/steps/oauth.py
@@ -4,11 +4,12 @@ import uuid
 from behave import given, then, when
 import requests
 
-from features.steps import utils
+from features.steps import s4s, utils
 from testsuite.oauth import authorize
 from testsuite import fhir
 
 
+AUTHORIZATION_ACTIONS = ['allow', 'deny']
 ERROR_AUTHORIZATION_FAILED = 'Authorization failed.'
 ERROR_BAD_CONFORMANCE = 'Could not parse conformance statement.'
 ERROR_OAUTH_DISABLED = 'OAuth is not enabled on this server.'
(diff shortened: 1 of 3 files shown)


features/steps/oauth.py
```diff
@@ -4,11 +4,12 @@
 from behave import given, then, when
 import requests
 
-from features.steps import utils
+from features.steps import s4s, utils
 from testsuite.oauth import authorize
 from testsuite import fhir
 
 
+AUTHORIZATION_ACTIONS = ['allow', 'deny']
```
erikwiffin: I'm using "allow" and "deny" instead of "approve" and "deny" because a) that was the language used in the allscripts `.feature`s and b) "approveing" is not a word.

features/steps/oauth.py
```diff
@@ -118,6 +119,37 @@ def step_impl(context):
...
+def step_impl(context, action, resource_type):
+
+    assert action in AUTHORIZATION_ACTIONS
+    assert resource_type in s4s.MU_CCDS_MAPPINGS
+
+    # Filter the steps to only what is required for this authorization
+    condition = '{0}.{1}'.format(action, resource_type)
+    steps = context.vendor_config['auth'].get('steps', [])
+    steps = [step for step in steps
+             if 'tag' not in step or step['tag'] == condition]
```
erikwiffin: I went with `tag` as the conditional name, but it doesn't quite feel right because it's not really behaving the way tags behave. I'd like to use something more indicative of the toggling behavior, (maybe `when`?), but I figured I'd run it by you before I made Allscripts rewrite their PR because I'm particular about how I name things.

jmandel: I agree that `when` is a clearer name!

erikwiffin:
Updated demo URL http://goo.gl/SJgJux

jmandel:
This looks like a great start! Given the current `.feature` files, it looks like we don't have any immediate need for a table-based listing of approve/deny decisions, right?

erikwiffin:
I don't think so, and I can't see it becoming necessary anytime soon either.

jmandel:
Great. Ready to merge this?

jmandel:
Hmm, wait -- the default authorizer is currently going to execute all of the steps automatically, huh? Really it should just execute the untagged steps, I think. Or perhaps the "approve" steps but never "deny" steps. Thoughts?

erikwiffin:
That's a really good point. I feel like it is more "correct" to only run steps without a "when" key, but since this is mostly a feature for Allscripts, it's probably worth checking with them to see if they actually plan on implementing an authorize screen that shows a big list of unchecked resources.

jmandel:
I believe that @jmblakl [said](https://github.com/sync-for-science/test-suite/pull/40):

> All boxes will be checked by default

So that covers the allscripts use case, at least. But let's play this forward: what if a vendor like Stpircslla presented a set of unchecked boxes. Then (even without introduction of `when` tags) they'd nee to supply a list of untagged "check the box" steps by default. And now (i.e. with this introduction of `when` tags) they'd add `When: deny.whatever`  tags to a new set of steps *below* the untagged ones.

So: I think I'm convinced that the correct default behavior is to filter out any tagged steps.

erikwiffin:
It sounds like the "allow" behavior isn't going to be terribly useful, since I can't think of any situation where you'd want to test allowing requests above and beyond the default, but it doesn't hurt anything to keep it around, and maybe someone will come up with something.

erikwiffin:
Don't merge yet, writing updated readme.

erikwiffin:
Ready on my end.

jmblakl:
Hey guys saw this come through I'm out of the office right now but will be back tomorrow.   In the meantime I'llI'lve Cody weigh in.

PR #51, opened 2016-07-21
Tagging updates and follow "next"

erikwiffin:
* Fixes https://github.com/sync-for-science/tracking/issues/11#issuecomment-234287037.
* Adds tags to individual SMART features so they can be tested independently.
* Adds an "optional" tag.

### DIFF
--- a/features/steps/fhir.py
+++ b/features/steps/fhir.py
@@ -21,9 +21,9 @@ def step_impl(context):
     urls = [link['url'] for link in links
             if link['relation'] == 'next']
 
-    assert len(urls) == 1, \
-        utils.bad_response_assert(context.response,
-                                  ERROR_NO_NEXT_LINK)
+    if len(urls) is not 1:
+        context.scenario.skip(reason=ERROR_NO_NEXT_LINK)
+        return
 
     context.response = utils.get_resource(context, urls[0])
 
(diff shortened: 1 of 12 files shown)


features/steps/fhir.py
```diff
@@ -9,9 +9,25 @@
...
 
+@when('I follow the "next" link')
+def step_impl(context):
+    resource = context.response.json()
+
+    links = resource.get('link', [])
+    urls = [link['url'] for link in links
+            if link['relation'] == 'next']
+
+    assert len(urls) == 1, \
```
jmandel: I don't think we want to go quite as far as requiring that there be a next link. But we do want to test the next link if it is present. The only failing condition should be a next link that is present but broken.

testsuite/oauth/authorize.py
```diff
@@ -156,7 +158,12 @@ def _execute_step(self, step):
         if not elem.is_displayed() and step.get('optional'):
             return
         elif not elem.is_displayed():
-            raise ElementNotFoundException('Element is hidden', self.browser)
+            try:
+                wait = WebDriverWait(self.browser, VISIBILITY_TIMEOUT)
```
jmandel: Did something change in one of the existing portals that made the previous approach break?

erikwiffin: Allscripts was failing as if I hadn't clicked the big "sign in" link. Changing the viewport size and adding this wait seemed to help. The old viewport size looked like a mobile view, that may have something to do with it. Another commit is coming that reworks this a little bit, so that it'll wait for optional elements too.

image

PR #52, opened 2016-07-22
Validate LOINC codes in lab results

erikwiffin:
First pass at this.

I think this UI works, but let me know what you think. Right now if a code fails to validate, the "validate" step will be marked as skipped, and there will be a note about it if whoever ran the tests wants to see it.

Not 100% sure the best way to include the various systems with this codebase. Right now I've just put loinc.csv in a data directory, but we're probably not actually allowed to do that/not a good idea anyways. I'm thinking that maybe we put some CSV files in an s3 bucket somewhere, and have the Dockerfile download them. That way we're at least not storing them in the github repo.

Also, UMLS has a kinda long and scary registration process, so I don't have snomed or rxnorm yet. Is there anything else we need?


jmandel:
Very cool! One approach I've taken in the past is to provide an external database with LOINC, SNOMED CT (the largest one, at 300k terms) and RxNorm (e.g., in a. Mongo db instance). Here, it would be good eventually  to explore the new HAPI support for vocabulary operations.

Yeah, working with UMLS can be a challenge -- even after you register and download all the files, there is a graphical user interface Java tool that is needed to actually extract the relevant vocabulary components. Some day it would be fun to  automate the process, if nobody has yet :-)

erikwiffin:
You don't still have that database running do you? :D

I can't find anything on the HAPI website about vocabulary operations, am I blind?

jmandel:
There actually is still a database running that supports the CCDA
Scorecard. It hasn't been updated to the latest releases of any of the
vocabularies, but for the record:

    mongo ds045757.mlab.com:45757/vocab -u smartread -p readonly

On the HAPI side, support for vocabularies (and load scripts for LOINC and
SNOMED) is a feature of the latest release; James Agnew advertised it on
the HAPI zulip stream. Worth investigating.

jmandel:
Neat! If you serialize the bloom filter after loading in SNOMED + LOINC, how big is this? In other words: what's the package size we could get away with (and the associate "false positive" rate)?

erikwiffin:
`codes.bf` is a 2.2M file (contains loinc and snomed), so that's not too bad. Bloom filters run pretty snappy, and eyeballing the logs, it looks like they're reporting accurately.

I'd probably delete `temp.py` before the final commit, but wanted to show it here so you can see where `codes.bf` is coming from.

I couldn't figure out which file `RxNorm_full_07052016.zip` actually contained the rxnorm list, so I'll get back to that.

As far as I can tell, if vendors are providing a system with their codes, the data is passing.

jmandel:
Re: accuracy, it would be good to confirm with a set of "negative examples" -- e.g. if you run a loop 1M times, generating a fake/random SNOMED codes, how often are they classified as valid?

erikwiffin:
That's a good idea. Fwiw, the library defaults to an error rate of 0.001 (https://github.com/jaybaird/python-bloomfilter/blob/master/pybloom/pybloom.py#L292) so that *should* be what we're looking at, but yeah, I'll write up some kind of negative example test.

erikwiffin:
It's a good thing I ran the test you suggested. I used a `ScalableBloomFilter` because it was supposed to manage the false positive rate better, but still going from 100 to 1 million records raised the rate from 0.001 to 0.008. Switched to 2 million and my test showed no false positives. Don't know why the default is so low, final file size is 4.6M.

erikwiffin:
I think the RxNorm list I'm using is missing deprecated codes. SMART fails the test for code 404673, it doesn't seem to exist here [https://rxnav.nlm.nih.gov/REST/rxcui/404673/ndcs.json] but does at [https://rxnav.nlm.nih.gov/REST/rxcui/404673/allndcs.json].

jmandel:
Yeah, https://rxnav.nlm.nih.gov/REST/rxcui/404673/status.json indicates it's been remapped. So this is a fair point: codes that aren't non-existent, but aren't current either.

I think the most conservative thing to do here is just try to load all the codes in rxnorm, regardless of deprecation status. That will help us focus on outright errors.

erikwiffin:
I haven't filtered out any deprecated codes, it looks like they're in a
separate file. Looks like I get those from
https://www.nlm.nih.gov/research/umls/sourcereleasedocs/current/RXNORM/sourcerepresentation.html#file9

erikwiffin:
RXNORM provides their deprecated codes in RXNCUI.RRF. The importer now reads both files and adds both of them to the bloom filter.

I noticed a problem with ICD-10 where `icd10cm_index.xsd` only contains the most specific codes (`I06.0`, `I06.1`, `I06.2`, `I06.8`, `I06.9`) but not the top level category (`I06`). Cerner is using the category in Problems, so I switched to importing `icd10cm_tabular.xsd` instead, which contains `I06`. Let me know if we want to tell them not to do that.

jmandel:
Gotcha -- thanks (I see this from [the docs](https://www.nlm.nih.gov/research/umls/rxnorm/docs/2012/rxnorm_doco_full_2012-3.html#s12_9) now that I look in the right place)

jmandel:
Re: ICD10, I think the advice is that if a category has more specific codes within it, then the bare category code shouldn't be used.

PR #54, opened 2016-07-28
Refactor sign in/authorize steps

erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/60.

Interestingly, no one but us seems to have a dedicated "authorize" button. It's pretty much just a "if you sign it, it's good" process.

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -17,10 +17,11 @@ auth:
   scope: launch/patient patient/*.read offline_access
   confidential_client:  # Should the test suite use basic auth while requesting tokens.
   token_url:  # Use when the token URL cannot be derived from a conformance statement
-  steps: []
+  sign_in_steps: []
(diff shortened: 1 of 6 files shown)


erikwiffin:
This led me to discover a weird bug, when flask is being served in development mode, it fails the `Special characters in "state"` scenario because the flask server decodes some of the special characters `\`[]\}^{` when returning `request.url`. The production uwsgi server does not have this bug. I'm going to log an issue with flask/werkzeug, but I wanted to leave a message here too to remind us to make sure we're always running the proxy server in production mode.

jmandel:
Interesting -- it's worth saying that allscripts has introduced a set of checkboxes as an authorization step, but once you click them (i.e. the first time you approve an app) that decision is remembered and they're not shown again.

PR #55, opened 2016-07-30
Ask for authorization overrides

erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/11#issuecomment-229064968.

### DIFF
--- a/features/20-exchange-access-token.feature
+++ b/features/20-exchange-access-token.feature
@@ -58,3 +58,11 @@ Feature: App exchanges authorization code for access token
             | key | value |
             | redirect_uri | https://example.com |
         Then the response code should not be 200
+
+    Scenario: Use received code twice
+        Given OAuth is enabled
+        And I am not logged in
+        When I ask for authorization
+        And I exchange my authorization code
+        And I exchange my authorization code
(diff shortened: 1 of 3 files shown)


jmandel:
For the "same code twice" issue, I'm expecting to see a `code` value being overridden, with a previously-supplied value re-supplied. Is this commit going through that workflow?

erikwiffin:
Oh, missed that line for some reason. Should be easy to add.

erikwiffin:
And this uncovered a bug in the `auth-proxy` codebase where we aren't properly expiring grants after they've been used. Looks like the expiration is being set, but the SQLAlchemy transaction isn't getting committed.

features/20-exchange-access-token.feature
```diff
@@ -58,3 +58,11 @@ Feature: App exchanges authorization code for access token
             | key | value |
             | redirect_uri | https://example.com |
         Then the response code should not be 200
+
+    Scenario: Use received code twice
+        Given OAuth is enabled
+        And I am not logged in
+        When I ask for authorization
+        And I exchange my authorization code
```
jmandel: I love that this is the test :)

PR #56, opened 2016-08-03
Add test complete report

erikwiffin:
This report doesn't indicate which systems we don't recognize, I wasn't sure how to do that without making us look silly since we aren't validation the various fhir valuesets, but it does list the systems that it finds, we can use that to identify that vendors are using custom systems or omitting them entirely.


erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/61.

jmandel:
Very cool! Just reviewing now -- some quick thoughts:

* would be good to avoid a modal message here, so the report doesn't disappear if you want to go look at test results in more detail. Maybe insert-and-scroll, or insert-then-add-button-that-scrolls

* For code report, would be good to highlight: how many codes from each vocab? Sort into a table with two cols (counts + code systems), sorted by count descending.

* For vocabularies where we can validate membership, would be good to add a %valid column, or "# invalid" column. 

* To improve the utility of this: we might also want to pre-load all FHIR valusets into our bloom filter. Details at http://hl7.org/fhir/valuesets.json (would want to grab all the valuesets that have a  "codeSystem" -- and then add the `ValueSet.codeSystem.concept` entries to that).

erikwiffin:
* There's a button that shows up at the bottom of the page to "show report" which re-opens the modal. I can move the report out of a modal, but I found that I liked having it there because it made it easy for me to tell when tests had finished running. What would you think of moving the button to somewhere more obvious?

* For bullets two and three, that's a good idea. Should be able to work that into this PR without too much trouble.

* Also a good idea, and I didn't realize there was a single json endpoint that returned all of the valuesets. Do you mind if I make a separate issue for that and work on it there?

jmandel:
Ah, I didn't notice the button in the UI, so yes, that'd help. Also if you keep the model, it might be good not to open it by default. My rationale: I was afraid to close the modal for fear I couldn't get it back. If I had clicked to open it, I wouldn't have had this fear. Alternatives could include some way to re-assure users that closing it was not a permanent dismissal.

For valueset incorporation: absolutely, feel free to create a separate issue for it. The one subtlety here is that we don't want to load *valusets* per se, but just to load the FHIR-defined code systems that appear in some valuesets. Happy to explain this in more detail if that's not clear ;-)

jmandel:
Also  we should track how many CodeableConcepts had **not** coding.

jmandel:
One more thought on UI for the report: may make sense for it to appear like the test result rows (e.g. just below "User revokes auth") -- not with the dots, but just a bit of text like "data vocabulary report", or somesuch, where clicking scrolls to content on the right.

PR #58, opened 2016-08-09
Add FHIR valuesets

erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/64.

The structure of http://hl7.org/fhir/valuesets.json is a little weird, (`codeSystem.concept.concept` for example) but I think I got everything.

That document doesn't contain the v2 (http://hl7.org/fhir/terminologies-v2.html) and v3 (http://hl7.org/fhir/terminologies-v3.html) valuesets, but I couldn't find a json file for either of those. Or is it better that we don't "recognize" them, and encourage people to use the fhir concepts when available?


data/build_bf.py
```diff
@@ -75,13 +77,50 @@ def import_icd10(bf):
...
+    systems = [entry['resource'] for entry in bundle['entry']
+               if 'codeSystem' in entry['resource']]
+
+    for system in systems:
+        url = system['codeSystem']['system']
+        concepts = system['codeSystem']['concept']
+
+        codes = []
+        for concept in concepts:
+            if 'concept' in concept:
```
jmandel: In theory this is a recursive structure that can go arbitrarily deep.

jmandel:
Thank for this tweak (don't know whether any built-in vocabs actually go three levels deeps, but it's nice when the "correct" code is more readable to boot!)

jmandel:
http://hl7.org/fhir/v3-codesystems.json

erikwiffin:
Fail the "All the codes are valid" test if codes do not match the system.

erikwiffin:
This also fixes https://github.com/sync-for-science/tracking/issues/67.

PR #60, opened 2016-08-10
Add footer

erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/63.


jmandel:
Thanks @erikwiffin! If we start listing DBMI and Fleet as orgs here the list is going to grow quickly (I just checked internally!). The word "partnership" can also be a tricky one in the instutitional context.

For the moment, could you adjust to just say something "This is the Syncfor.science test suite. Source code at [[GitHub]]"? (And can work in the background on how to include a shout-out to Fleet.)

jmandel:
Thanks for tweaking!

PR #62, opened 2016-08-16
Validate Argonaut profiles

erikwiffin:
This adds a validator to the Argonaut smoking profile. Wanted to get your feedback before I set it up with everything else. This approach unfortunately fails on the first error it sees, but (in an unusual instance of actually using behave the way it's intended to be) I was able to use a lot of the language that already existed http://argonautwiki.hl7.org/index.php?title=Smoking_Status#Complete_Summary_of_the_Mandatory_Requirements.

I tried the standalone validator http://hl7.org/fhir/validation.html#jar and the HAPI validation http://hl7.org/fhir/validation.html#op, but couldn't get either of them to validate profiles.


jmandel:
I think the logic here is quite clear / readable! Regarding behavior though: the intention was for the "Argonaut profile checks" to show up as data quality indicators at the end of the report, without causing vendors to fail if they're wrong. Any thoughts about how to make this work, or how to otherwise provide this kind of feedback without outright failing for non-Argonaut-conformance?

Also regarding the validation JAR + HAPI: it would be good to document what you tried, since these should (in theory) work somewhat. Did you encounter failures, or "success" responses that didn't detect errors?

erikwiffin:
Good point. I forgot about that. I'll play around with it a bit, I might be able to pull it into the `before_feature` in `environment.py`, and grab that output and attach it to the report.

As for the jar, I basically just followed the exact steps in the documentation (but in bash, instead of what I assume is dos). I got validation errors for invalid FHIR resources (ex: unrecognized fields) but wasn't able to trigger any errors for situations where I failed to conform to the specific profiles (ex: leaving off `Patient.identifier.use` or providing a `Patient.contact.address.postalCode` that was clearly not a valid zipcode).

jmandel:
Thanks! And what happened with HAPI?

erikwiffin:
Same thing. I think there might be a "load the profile into HAPI" step that I was missing, but the documentation on their site involved writing custom profiles in Java, and rebuilding the server, so I'm pretty sure that's not what we want.

I have a pet theory that the broken redirect I told you about yesterday is part of the problem, but not sure how to test that.

erikwiffin:
More to come, just wanted to get a checkpoint up onto github.

erikwiffin:
This is about it, but there are some gaps (I think I left comments for most of them) where actually validating the profile would get more difficult.

jmandel:
Comments in the feature files you mean, like the following?

        # And DocumentReference.content.attachment.contentType is bound to http://www.rfc-editor.org/bcp/bcp13.txt

It would be good to make sure we're capturing *all non-validated Argonaut Profile expectations*, rather than just most of them :-) In part because things that are hard for us to validate are going to be hard for others, too -- and the community should either build tooling, or eliminate some of these constraints.

---

Some issues with the following, I think:

```
+        for code in found:
 +            coding = {
 +                'system': value_set_url,
 +                'code': code,
 +            }
 +            valid = systems.validate_coding(coding)
```

It looks like this is putting the `value_set_url` inside of a `Coding.system` -- but this is mixing up the "value set" with the "code system". Easiest to talk about this with an example:

Argonaut says smoking status is bound to "http://hl7.org/fhir/ValueSet/daf-observation-ccdasmokingstatus". But instance data wouldn't mention `http://hl7.org/fhir/ValueSet/daf-observation-ccdasmokingstatus` -- they'd look like:

```
{
  system: "http://snomed.info/sct",
  code: "449868002"
}
```

or 


```
{
  system: "http://hl7.org/fhir/data-absent-reason",
  code: "not-asked"
}
```

 -- in other words, there's one level of indirection missing from these tests. 

I'm also not seeing validation of `Observation.code` for lab tests, or drug codes, or problem codes (possibly I'm just missing where this occurs?)

features/00-lab-results.feature
```diff
@@ -16,31 +16,36 @@ Feature: Lab results
...
     Scenario: Resources fulfill the Argonaut Laboratory Results profile
         Given I have a Lab results response
+        And there is at least 1 entry
         Then there exists one status in Observation.status
         Then there exists one category in Observation.category
         And there exists a fixed Observation.category.coding.system=http://hl7.org/fhir/observation-category
         And there exists a fixed Observation.category.coding.code=laboratory
         Then there exists one code in Observation.code
         And there exists a fixed Observation.code.coding.system=http://loinc.org
-        And Observation.code.coding.code is bound to http://loinc.org/
```
erikwiffin: No more silently passing. Yay!

erikwiffin:
This fixes a lot of what we talked about, but there are a few more things in the comments.

* Added a "recognized" column to the report. As per our discussion, this only recognizes systems, not value sets. Although a lot of things are actually coded with value sets (in the SMART data at least), so that might be worth revisiting?
* Included http://argonautwiki.hl7.org/index.php?%20title=Argonaut_Vital_Signs in the bloom filter. Ironically, one of the SMART vital sings was coded with 3141-9, which is excluded.
* New "given" that skips tests if there isn't at least 1 entry. Should in more obvious test results.
* Reworked the "validate" logic. Two functions now for validating both dictionaries of codes/systems, and codes against a valueset/system. Also raises an exception now if the system is not found.

One outstanding thing is that I was going to import the CPT codes for Procedures, but I can't find them. This https://www.cms.gov/medicare/fraud-and-abuse/physicianselfreferral/list_of_codes.html was the best I could find and I don't think it's the whole list.

jmandel:
Thanks @erikwiffin and sorry for the slow response! 

> Although a lot of things are actually coded with value sets

Could you explain what you mean (maybe with an example or two)? We might be using different words to refer to the same things. 

Regarding CPT, I think we should be able to do probabilistic testing without having to redistribute these codes; can you download the UMLS release from https://www.nlm.nih.gov/research/umls/licensedcontent/umlsknowledgesources.html (should include CPT, per https://www.nlm.nih.gov/research/umls/sourcereleasedocs/current/CPT/)

erikwiffin:
No worries, hope you're enjoying your vacation!

image

Based on what I can remember of our conversation a while back, things like `http://hl7.org/fhir/observation-category` are considered value sets, not systems, and they don't count as "recognized". These show up in Observation.category, in our vital signs records.

Thanks for pointing me at the CPT codes. I'll download that and let you know how it goes.

jmandel:
Thanks! Time off was nice -- this week I'll be in "Noogler orientation" at Verily. 


Regarding value sets: the screen shot looks like it's conveying exactly the right information. In FHIR terms, http://hl7.org/fhir/ValueSet/observation-category is a value set that defines a code system (http://hl7.org/fhir/observation-category).

So when a structure definition binds Observation.category to this value set, it means that instance data should have a category code with system "http://hl7.org/fhir/observation-category". So this particular value set *defines a system*.  Instance data only ever refer to the system. Does that make sense?

erikwiffin:
I think so. It's a very confusing set of terminology. :)

I'll take another look at the FHIR valueset document, because I think that _ideally_ we don't have a bunch of "unrecognized" x's next to things coming out of the FHIR spec.

jmandel:
Yes -- all systems define in the FHIR spec should be available from the terminology json, and we should be extracting them. I think your current approach (the recursive walk) is nearly right; you just have to associate the codes you find with the correct (FHIR defined) system URI.

PR #67, opened 2016-09-21
Update athenahealth.yml

kcheng3:
We're making a switch from testing api (Mashery) to apitest, which is simpler for us to test


jmandel:
Addressed elsewhere!

PR #69, opened 2016-10-14
Add OAuth flow to eClinicalWorks

erikwiffin:


testsuite/oauth/authorize.py
```diff
@@ -99,7 +99,12 @@ def _browser(self):
         """
         RemoteConnection.set_timeout(CONNECTION_TIMEOUT)
 
-        return webdriver.Firefox()
+        profile = webdriver.FirefoxProfile()
+        profile.set_preference('general.useragent.override', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.116 Safari/537.36')  # noqa
```
jmandel: Can you add a comment explaining this? And does this override work with all the other vendors? Should we extarct this out so it can be configured on a per-vendor basis?

testsuite/oauth/authorize.py
```diff
@@ -99,7 +99,12 @@ def _browser(self):
...
+        driver = webdriver.Firefox(profile)
+        driver.implicitly_wait(1)
```
jmandel: Also, can you comment on this wait? What does it do?

testsuite/oauth/authorize.py
```diff
@@ -184,10 +189,12 @@ def provide_user_input(self):
 
         steps = itertools.chain(self.config.get('sign_in_steps', []),
                                 self.config.get('authorize_steps', []))
+
         for step in steps:
             self.runner.execute_step(step)
 
-        query = self.runner.get_query(base_url=self.config['redirect_uri'])
+        base_url = self.config['redirect_uri'].replace('http://', '')
```
jmandel: This is to deal with the redirection-adds-an-`s` issue?

testsuite/oauth/authorize.py
erikwiffin: It looks like it works for all the other vendors, but we probably should pull it out into a config option. I'm a little sad that it's even necessary, so formalizing it into a config option feels like a painful concession, but that's still probably the way to go.

PR #72, opened 2016-10-26
Test "do not authorize" flow

erikwiffin:
This is incomplete, as Allscripts appears to be down at the moment, so I can't update their config.

I wanted to get your thoughts on my implementation though. When the "I abort an authorization request" step is reached, it replaces the "authorize_steps" part of the config with "cancel_steps". This worked well for SMART and eClinicalWorks, who already had a distinct "authorize" flow, but some of the vendors treat logging in as authorizing, so I had to break the "clicking the login button" step into an "authorize" flow. I don't think that's quite correct, but it doesn't cause any problems. Thoughts?

### DIFF
--- a/config/cerner.yml
+++ b/config/cerner.yml
@@ -13,7 +13,11 @@ auth:
       send_keys: "ARGONAUT"
     - element: "#j_password"
       send_keys: "sprinttest"
+  authorize_steps:
     - element: "#loginButton"
+      click: True
+  cancel_steps:
+    - element: "#cancelButton"
       click: True
   extra_launch_params:
     launch: 8de5dad5-4e22-4b06-bde5-60f6a5781fd0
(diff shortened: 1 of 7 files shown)


jmandel:
I think this is the right framework -- though when it comes to the allscripts behavior, I believe that the *first* time a user authorizes the app, they get a more detailed approval flow (which our client currently isn't configured to handle). That's the place where a "cancel" would be applicable. I think it's hard for us to test this automatically though, given that authorization decisions are remembered once made. @jmblaki does this make sense? Is there any programmatic way to ensure that our authorization request is treated like a first-time request?

jmblakl:
@jmandel there isn't today, but we can add that.  It may be a sprint or two because we are feverishly trying to finish work before out last release of the year.

erikwiffin:
Since there is currently no way to cancel the Allscripts re-authorization, I think this is ready to merge.

Fixes https://github.com/sync-for-science/tracking/issues/74.

PR #73, opened 2016-11-04
Click the drchrono authorize button if presented

erikwiffin:
Drchrono seems to "remember" authorizations, so the authorize screen is not always presented. Otherwise, the "cancel_steps" could also be implemented.

### DIFF
--- a/config/drchrono.yml
+++ b/config/drchrono.yml
@@ -16,3 +16,7 @@ auth:
       send_keys: "X6mx5AYmT7rUdVv44sZP"
     - element: "button[type='submit']"
       click: True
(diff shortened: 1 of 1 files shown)


saq7:
Yeah, it does have a somewhat long lived token at the moment. We could change that, but it would be great if the test suite could handle both scenarios. Though if it is problematic, let me know.

erikwiffin:
I've configured the authorize step as optional, so it's not going to fail any tests as is, but if we wanted to be able to test the "cancel" path, then we'd need some way of ensuring that the authorize page is displayed consistently.

PR #74, opened 2016-11-07
Recognize Argonaut Extensions codes

erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/78.

Are there any known uses of the argonaut extension codes so we can see this working?

### DIFF
--- a/data/.gitignore
+++ b/data/.gitignore
@@ -12,3 +12,4 @@
 !fhir/daf.json
 !fhir/argo.json
 !fhir/argo-vital-signs.json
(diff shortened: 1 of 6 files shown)


jmandel:
The Athena server is using some. That's actually where I first saw them.

erikwiffin:
Ironically, they are using the wrong system. "http://argonaut.hl7.org/extension-codes" vs. "http://argonautwiki.hl7.org/extension-codes".

image

jmandel:
Ironically indeed -- and good call! I'll send them a note :-)

PR #75, opened 2016-11-14
Feature/save testrun

erikwiffin:
Instead of trying to shim a report view on it's own into the display, I decided it would be easies to just use the existing infrastructure and save the entire test run. Old runs can be reconstituted as if they were just a new snapshot. http://tests.dev.syncfor.science:9003/?report_id=4 was a test I ran against epic.

Fixes https://github.com/sync-for-science/tracking/issues/77.


erikwiffin:
1) Since I'm storing the plan and the snapshot, and the JS doesn't know anything about our actual featureset, it should be remarkably resilient to adding new tests and features. Lucky thing too, because I hadn't considered that at all :)
2) As long as the TestRun model has an `event` property that provides a properly formatted dictionary object, the actual table schema shouldn't matter at all.
3) If we create a volume for the `db.sqlite3` file, like our other projects, it should be able to survive a docker restart.

My only concern is that the database will grow pretty quickly if people are doing a lot of testing. Maybe Tom can wire this one up to an external sql database too :)

jmandel:
That's fair. I think we can try it and keep an eye on disk space. I wonder if we need to be writing each chunk to the database as they arrive, vs just at the end. But I may be prematurely optimizing... is there a use case for having partial results stored in the db? 

The alternative I was thinking about would just be dumping static reports to a volume as html files, like we do with screenshots. In some ways it's more robust, but I really like that your approach preserves the whole UI.

erikwiffin:
Because of the way I have the sqlalchemy model defined, each partial result per test run overwrites the previous one. The end result being that we only have... the end result.

jmandel:
Yeah no, I got that. I'm just wondering if there's a use case for having partial results stored in the db,  or that's just a side effect.

erikwiffin:
Well, if for some reason the test suite crashes halfway through a run, we'd have the partial run stored as well. We'd lose the error message though.

testsuite/models/testrun.py
```diff
@@ -0,0 +1,80 @@
+''' Test run module.
+'''
+import json
+
+from testsuite.extensions import db
+
+
+class TestRun(db.Model):
+    ''' The results of a single test run.
```
jmandel: It looks like you could maintain a normalized db with just one table containing columns for plan and snapshot. I don't necessarily object to having 3 tables but I'd like to understand your rationale.

testsuite/tasks.py
```diff
@@ -1,44 +1,53 @@
...
+@celery.task
+def run_tests(room, vendor, tags, override):
+    app = create_app()
+    with app.app_context():
+        test_run = TestRun()
+        db.session.add(test_run)
+        db.session.commit()
+
+        def on_snapshot(snapshot, plan):
+            test_run.save_snapshot(snapshot, plan)
```
jmandel: Do we expect this should ever fail? Should we be defensive and let tests continue even if there's an error here?

testsuite/models/testrun.py
jmandel: Also now that we've got a  db here it'd be great to capture date, vendor, and some kind of "is a complete test run including all tags" boolean at this stage. 

FEATURE CREEP!

(But let's keep those in mind so we can facilitate reporting, which is currently a manual process.)

testsuite/views.py
```diff
@@ -1,37 +1,61 @@
...
+@app.route('/load-report/<int:test_run_id>', methods=['POST'])
+def report(test_run_id):
+    room = request.form['room']
+    run = TestRun.query.get(test_run_id)
+    event = run.event
 
-    @socketio.on('connect')
-    def cb_handle_connect():
-        flask_socketio.send('connected')
+    socketio.emit('snapshot', event, room=room)
```
jmandel: Pretty awesome that you can just emit the old result this way :)

testsuite/models/testrun.py
erikwiffin: I don't have a good reason for putting them in separate tables, other than they *feel* like distinct objects. In particular snapshots, if we ever decided to start storing partial runs, that would be a pretty minor change at this point.

As for feature creep, I'm also planning on using this same db for https://github.com/sync-for-science/tracking/issues/76. I'm hoping that instead of passing the output along in a POST param, I'll be able to just pull it out of the database. Not 100% sure how it'll work yet.

testsuite/tasks.py
erikwiffin: It could fail if something in the JSON were unserialize-able. That can happen with `datetime` objects, but if we've gotten this far, everything should already be safe (and if it isn't, `socketio.emit` will fail).

I think, because of the non-critical nature of this app, I'd rather have everything fail, post an error message, and have the vendors complain to us, rather than making sure the tests run at any cost, and possibly silently losing "save report" functionality.

testsuite/views.py
erikwiffin: Yeah, I was a little bit dreading figuring out how to re-render the report until I realized I could do this.

jmandel:
Cool, and thanks for the instantaneous response :

This works for me

PR #82, opened 2016-12-05
Improve error reporting

erikwiffin:
- Earlier invalid YAML detection https://github.com/sync-for-science/tracking/issues/86
- Display bad conformance statement response https://github.com/sync-for-science/tracking/issues/85


jmandel:
Thanks @erikwiffin !

PR #85, opened 2017-01-09
New Config value for fhir api

jmblakl:
just changed the api url.  both are still live urls, but we want to destroy the old soon.


jmandel:
Thanks @jmblakl! @erikwiffin will you merge and help propagate this to our deployment configs?

PR #86, opened 2017-01-09
Feature/revocation

erikwiffin:
Tests Allscripts revocation. Adds some of the config for Epic, but our credentials aren't working for now, so keep skipping.

Also fixes Allscripts authorize bug.

### DIFF
--- a/config/allscripts.yml
+++ b/config/allscripts.yml
@@ -20,3 +20,13 @@ auth:
       send_keys: "s4s!2345"
     - element: "[translate='Login_LogIn']"
       click: True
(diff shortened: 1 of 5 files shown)


jmandel:
Nice! Should I be worried about the fact that this is failing? Also, I just discovered why GitHub has the funny rule that you can't merge a failing pull request from a phone: I accidentally pressed the merge button with my finger and I was thrilled that it didn't go through.

erikwiffin:
From "why would you want that" to "oh that's why" in how long? :D

PR #90, opened 2017-01-27
Wait for Epic login page to load

erikwiffin:


config/epic.yml
```diff
@@ -13,6 +13,7 @@ auth:
   sign_in_steps:
     - element: "#Login"
       send_keys: "fhirjason"
+      wait: 5
```
jmandel: Can you help me understand this wait? Normally I thought we wait (until a default timeout) for every element to appear before proceeding.

erikwiffin: There's an "implicit wait" of 1 second for every element https://github.com/sync-for-science/test-suite/blob/master/testsuite/oauth/authorize.py#L127, but setting a wait in the config adds a hard wait in addition to that https://github.com/sync-for-science/test-suite/blob/master/testsuite/oauth/authorize.py#L58.

PR #94, opened 2017-02-05
use BASE_URL if available

jmandel:


erikwiffin:
Is this related to the local stack you're working on? From what I can tell, the `tests` service shouldn't have `BASE_URL` defined https://github.com/sync-for-science/reference-stack-docker/blob/master/deploys/demo.syncfor.science/docker-compose.yml#L54

jmandel:
Yeah, bad on me for not explaining.

Yes, it's related to the `pilot-prep` stack, where there's no letsencrypt-enabled proxy in the loop. I like the pattern of just being able to tell an app "this your canonical URL" (which we do currently for our `smart` portal component).

erikwiffin:
That's fair. If we do that though, I think I'd want to push for getting rid of the rest of that conditional, so that BASE_URL is the *only* way of determining a canonical URL. That make sense?

jmandel:
Yes, that makes sense. I just didn't want to break our existing config.

jmandel:
Realize we never got this merged -- do you prefer to remove the conditional and rewrite our existing docker-compose files to add `BASE_URL`?

PR #95, opened 2017-02-07
Insert dot into ICD9 codes for validation

erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/95

I tried to find an alternate source of ICD9 codes, hopefully something that already had the dots in them. Apparently no one is offering that. This just splices the dot in at position 3 and adds that (as well as the un-dotted code) to the BF.


jmandel:
Thanks @erikwiffin! Could you update to require *just the dotted form*?

erikwiffin:
Done. Why are dots required if the list is never distributed with them? Or, why is the list never distributed with them if they are required?

jmandel:
Thanks. I wish I could answer those questions with something better than "I have no idea" :/

PR #98, opened 2017-02-14
updated config file

jmblakl:
This reflects the pilot phase environment urls, as well as the new user with better data.  It does not include changes for the workflow bug discussed in today's (2/14/17) meeting.


jmandel:
Thanks @jmblakl! @erikwiffin will you merge this here (and into our deploy config)?

PR #99, opened 2017-02-20
Wait the appropriate amount of time when testing Allscripts

erikwiffin:
- Clear the report id when starting a new test run, so that tests that fail before a report id are generated don't pull up the old report.
- Fix the waits in the Allscripts config.
- Blindly accept any alerts that happen at any time in selenium.


erikwiffin:
I don't know if this should actually get merged. It was very much a case of hitting the code with a hammer until the tests pass. I was able to make it through two sequential runs against the Allscripts demo server with "App uses a refresh token to obtain a new access token" and "User revokes authorization" passing.

The alerts stuff was added because my dev setup doesn't use https, and Allscripts has an alert warning the user about that.

jmandel:
This doesn't look as bad as I feared from your comment — it's effectively adding a way before clicking the approve buttons (plus the alert stuff — which perhaps we should strip out first, if this was really a debugging issues in your dev enviornment, where perhaps just using https would be a better way to avoid special cases).

Have you seen this fail since adding the waits?

Let's talk about the "correct" way to do this, so we can make an informed decisions about the pragmatics of these "wait"s. I think the "correct way" would be:

1. Add support for branching logic in our authorize_steps
2. Use branching logic to determine whether we're in an "already approved" state vs a "need approval state"
3a. In "already approved" state, just wait until the redirect happens (no steps to do)
3b. In the "need approval" state, click the approve buttons in sequence as non-optional steps (which will add implicit waits for the elements to appear, and will let us remove the explicit waits that we've hacked in)

Does this sound right?

erikwiffin:
Ran the tests again, still passed. We might have found just the right amount of waits.

I re-jiggered the alert stuff, now it wont recursively automatically accept everything, all the time, but should handle them and error out more informatively.

erikwiffin:
Regarding your "correct way". I agree in theory, but adding support for branching logic... I don't want to make the config files a lot more complicated, and I can't really think of a good way to make this work without doing that.

jmandel:
Thanks - and agreed, if we can get by without turning our config files into a Turing complete programming language, we should :)

PR #100, opened 2017-03-03
Update Cerner for patient launch and new server

daliboz:
Updated the Cerner configuration to use a confidential client, use patient sign in and patient/launch, and added the revoke URL configuration.

I cannot figure out how to use the existing configuration structure to automate the revoke. Right now, a patient can revoke per refresh token, and the selector would use value of (all) the refresh tokens that need to be revoked.


jmandel:
Thanks @daliboz!

@erikwiffin, confirming that you will take care of this merge and the associated updates?

PR #104, opened 2017-03-15
Get list of configs from config directory

erikwiffin:
Dynamically generate the dropdown of possible configs to make testing the pilot prep stuff easier. Has the downside of removing any custom capitalization for names like "McKesson". Also, the list is now alphabetically sorted, which means SMART isn't at the top, and "Other" isn't at the bottom.

How important do you think those issues are?


jmandel:
I like this! Could we just use the base filenames with their on-disk capitalization and spacing (so `SMART` and `McKesson` stay intact, and in the future we can add clinic-specific entries as needed)?

As for the alpha sort: I'd kind of like to keep "Other" at the bottom, which would be easy enough to filter out of the list and hard-code in HTML... SMART can default to alpha-sort. What do you think?

erikwiffin:
So you're suggesting renaming `smart.yml` to `SMART.yml`?

jmandel:
Yeah, and McKesson, and any relevant clinic names for our internal pilot testing.

jmandel:
Great -- ready to merge?

erikwiffin:
👍

PR #105, opened 2017-03-22
De-inline large HTTP payloads

erikwiffin:
This should fix both of the payload related issues at once.

* HTTP responses are now collected and displayed before the test output.
* Each line of error output is truncated at 80 characters.
* I also upgraded webpack to v2, and split app.js into app.bundle.js and vendor.bundle.js.


jmandel:
Thanks @erikwiffin! This is a great start. A few notes:

 * in the current implementation, payloads appear on first run but disappear when re-loading a report. I understand that you're not storing these in the "snapshot", but perhaps they could still be stored in the db and replayed when the page is reloaded? Ideally, loading the page and re-loading the page should produce the same result. 

 * It would be good to preset (links to) payloads alongside the tests that generated them, rather than all at the top. I recognized that having the context for this may be challenging "in general" but if necessary, we could even re-establish the links client-side by inspecting the payload URLs.

 * truncated error responses scroll quickly (good!) but make some data invisible. Can we make it so there's a button like "open truncated error message in new tab"?

assets/main.js
```diff
@@ -60,6 +61,10 @@ $(function () {
       .html(summary_tmpl({summary: summaryResult.summary}))
       .find('[data-toggle="tooltip"]').tooltip();
 
+    $('#payloads')
+      .html(payloads_tmpl({payloads: payloads}))
+      .find('[data-toggle="tooltip"]').tooltip();
```
jmandel: Where is this tooltip in the UI?

jmandel:
Also for errors: would be good to replace the whole inline payloads with an "open in new tab" link, as these can be quite long vertically (as well as horizontally)

erikwiffin:
* Storing every single successful payload would add a pretty significant amount of data to the database. Are you sure we want to do that?
* I can take a closer look at that, but  yeah, establishing context for these is tricky. I can't just associate URLs, because some of the tests involve clearing the cache and making the same request again, with different login credentials.
* I'll just replace the entire inline payload with a link to open in a new tab. That way we don't have to worry about vertical/horizontal truncation.

There's no tooltip in the UI, I just wanted to be consistent with the previous two template renderings. If we add a tooltip to the payload template, it would be nice for it to just work, rather than needing to figure out what's different about that template vs. the other two.

jmandel:
I have mixed feelings on saving payloads; for our pilot testing we'll be persisting these to the db behind our "Demo Research App", so we don't have to worry about *losing* them...

For inlining the responses, couldn't we just tack onto https://github.com/sync-for-science/test-suite/pull/105/files#diff-0f04c2be6a879f3dc6e7dc6bd738b7c8R72 and read `context.feature` and `context.scenario` and pass them as args like:

        context.config.on_payload(url, bad_response_assert(response, ''), context.feature, context.scenario)

erikwiffin:
I've moved the responses into the features that generated them. I think it makes more sense here than next to scenarios, since we're caching the result, and we'd only make it once. We'd either duplicate every single scenario, or only the first one would have it, and neither of those sounded good to me.

Error outputs are now also in a new tab.

jmandel:
Nice -- I think that's a great place to fit them. Any other changes you want to incorporate before we merge?

erikwiffin:
I think I'm happy with it. Merge it!

PR #110, opened 2017-08-03
Only check match entries when counting results

erikwiffin:


features/steps/fhir.py
```diff
@@ -144,7 +145,8 @@ def step_impl(context):
 @given('there is at least 1 entry')
 def step_impl(context):
     resource = context.response.json()
-    entries = resource.get('entry', [])
+    entries = [entry['resource'] for entry in resource.get('entry', [])
```
jmandel: Technically this property is optional, so I am thinking that a more precise logical assessment could be to look for resources of the appropriate type, with either a mode=match or an empty mode.

erikwiffin: I'm not checking entry type, because I'm pretty sure returning just a Condition result in an Observation query would raise an error somewhere else, but `entry.get('search', {}).get('mode', 'match')` can be read as "get the search property of the entry, defaulting to an empty object if not found, then get the mode property of that, defaulting to match if not found". Should result in returning "match" if the property was left off (which =='s match) or the actual value if present.

jmandel: Indeed -- I get `get`'s behavior :-) The scenario I'm concerned about is if we run `GET /AllergyInteolerance?patient=123` and we receive bundle with a single result, an `OperationOutcome` with no `mode`. The issue with relaxed checks here is that we tend never to notice when we've got them wrong (case in point with this issue, which I think has been present for many months but it leads to green check marks in the test suite, which makes everyone happy ;-))

erikwiffin: Hmm, that makes sense. Should I only update the two tests I've got here to do that? And the logic they use to know what resources to filter out... Should I change the test to `there is at least 1 {resource_type} entry`?

jmandel: Thanks! Yes, I'd start with these two tests, since they give us insight into which sample data might be missing. (Let me know if you think there are other important targets!)

I think making the logic explicit in the feature language would be perfect and straightforward (at the expense of having to update 10ish individual feature files).

PR #111, opened 2017-08-03
Add a 'run all tests' endpoint and workaround btoa issue

jmandel:
This isn't exactly a "good idea"... but it's a start toward lowering the level of effort required to trigger a bunch of tests all at once.

### DIFF
--- a/assets/templates/base64.js
+++ b/assets/templates/base64.js
@@ -1,5 +1,9 @@
 var Handlebars = require('handlebars');
 
 module.exports = function (message) {
-  return btoa(message);
+  try {
+    return btoa(message);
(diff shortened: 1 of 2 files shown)


erikwiffin:
What exactly is the point of "headless_begin"? Does it dump all the test results out to the cli logs?

jmandel:
It's expected to write results to the db.. which we could later query and generate reports over.

erikwiffin:
That makes sense. Would be nice if the json blob you're returning contained all the test run ids, so that it would be easier to pull them up later, but I can't think of a good way to do that the way things are structured.

If you fix the pep8 issue, I'll merge.

PR #112, opened 2017-09-28
Py 3.6, new deps

jmandel:
This was a long-standing issue of questionable importance. After banging my head against it for a while: this works (for me... I think) but it involved things like inserting a magic incantation of 

```
import grequests
import requests
```

into the celery and flask app to prevent https://github.com/kennethreitz/grequests/issues/103 .

Net win? Unclear. Thoughts @erikwiffin  ?


jmandel:
Note that each of the code file changes is "necessary" in the sense that things stop working if you reverse them.

PR #115, opened 2017-10-06
sync-for-science/tracking#120 - Adding new test for values in Observa…

mtmcduffie:
…tion or a reason for absence.

### DIFF
--- a/features/steps/argonaut.py
+++ b/features/steps/argonaut.py
@@ -1,10 +1,11 @@
 # pylint: disable=missing-docstring,function-redefined
-from functools import reduce
+
 import json
 
 from behave import then
 
 from features.steps import utils
+
 from testsuite import systems
 
 ERROR_CODING_MISSING = '''
@@ -22,18 +23,10 @@ ERROR_INVALID_BINDING = '''
 '''
 ERROR_REFERENCE_MATCH = '{reference} is not a {resource_type}.'
 ERROR_REQUIRED = '{name} not found.'
+ERROR_MISSING_FIELD = ('The resources identified by ids ({resource_ids}) were absent'
+                       ' all of the following fields: {field_list}')
+
 ERROR_WRONG_FIXED = 'None of {values} match {value}.'
-
-
-def traverse(resource, path):
-    def walk(data, k):
-        if isinstance(data, dict):
-            return data.get(k)
-        elif isinstance(data, list):
-            return [reduce(walk, [k], el) for el in data]
-        return None
-
-    return reduce(walk, path, resource)
 
 
 def get_resources(resource, filter_type):
@@ -58,7 +51,7 @@ def step_impl(context, name, field_name):
     resources = get_resources(context.response.json(), filter_type)
 
     for res in resources:
-        found = traverse(res, path)
+        found = utils.traverse(res, path)
         assert len(found) > 0, utils.bad_response_assert(context.response,
                                                          ERROR_REQUIRED,
                                                          name=name)
@@ -73,9 +66,9 @@ def step_impl(context, field_name, sub_field):
     resources = get_resources(context.response.json(), filter_type)
 
     for res in resources:
-        found = traverse(res, path)
+        found = utils.traverse(res, path)
         for item in found:
-            match = traverse(item, sub_path)
+            match = utils.traverse(item, sub_path)
             assert match is not None, \
                 utils.bad_response_assert(context.response,
                                           ERROR_REQUIRED,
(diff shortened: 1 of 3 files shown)


mtmcduffie:
Sorry for multiple PR Requests. I was trying to figure out the syntax for GitHub's inter-repo issue tagging.

features/steps/argonaut.py
```diff
@@ -100,6 +100,25 @@ def step_impl(context, resource_type, field_name):
                                       resource_type=resource_type)
 
 
+@then(u'one of the following paths exist {field_string} in {resource}')
```
jmandel: Nit: can you add a colon after exist?

features/steps/argonaut.py
```diff
@@ -100,6 +100,25 @@ def step_impl(context, resource_type, field_name):
...
+def step_impl(context, field_string, resource):
+
+    fields_to_find = field_string.split(",")
+
+    resources = get_resources(context.response.json(), resource)
+
+    found = False
```
jmandel: I think this initialization step needs to happen inside the loop to get the behavior you want. Otherwise one good resource will shadow a pile of bad ones.

Another way to skin this logic would be something like the following (untested):

```py
def some_field_exists(resource, fields):
    return bool([traverse(resource, field) for field in fields])

def step_impl(context, field_string, resource):
    fields_to_find = field_string.split(",")
    resources = get_resources(context.response.json(), resource)
    valid_resource_ids = set([
        res.id for res in resources
        if some_field_exists(res, fields_to_find)])

    all_resource_ids = set([res.id for res in resources])
    invalid_resource_ids = all_resource_ids - valid_resource_ids
    assert len(invalid_resource_ids) == 0, \
        utils.bad_response_assert( # include the list of invalid resource ids here


```

features/steps/argonaut.py
mtmcduffie: Done!

features/steps/argonaut.py
mtmcduffie: Good catch. I implemented as you advised, spent some time banging my head on the reduce function when it didn't work like I expected. Thanks for the direction!

Error output should look like

Assertion Failed: The resources identified by ids (smart-observation-828-weight, smart-observation-828-oxygensaturation, smart-observation-828-temperature, smart-observation-828-heartrate, smart-observation-828-respiratoryrate, smart-observation-828-bmi, smart-observation-828-systolic, smart-observation-828-diastolic, smart-observation-828-height, smart-observation-828-bp) were absentall of the following fields: valueQuantity.value,component.valueQuantity.value,DataAbsentReason

features/steps/utils.py
```diff
@@ -139,6 +141,30 @@ def _clean(loggable):
...
+    def walk(data, k):
+        if isinstance(data, dict):
+            return data.get(k)
+        elif isinstance(data, list):
+            return [reduce(walk, [k], el) for el in data]
+        return None
+    return reduce(walk, path, resource)
+
+
+def is_field_in_resource(resource, fields):
```
jmandel: Naming nit: singular field in name vs are at argument. Maybe call the `some_field_present` or `any_field_present` or `has_one_of`?

features/steps/utils.py
```diff
@@ -139,6 +141,30 @@ def _clean(loggable):
...
+    """
+    Explores a dictionary looking for the existence of fields, can take "."
+     separated hierarchical representation of individual fields.
+    :param resource: dict
```
jmandel: Are these dicts or fhir clients objects?

mtmcduffie:
While we're here. A lot of this is dangerously close to a generalized JSON validator, if we were to have the Argonaut or other schemas to validate against. I didn't see any super promising JSON validation frameworks out there, have you guys run into one or considered the possibility?

jmandel:
There's json schema, which is reasonably expressive. In the other direction,  fhir also has its own language (StructureDefinition) for defining these constraints, with a Java implementation for running checks that can be invoked through HAPI.

jmandel:
Thanks for making these tweaks! Looks good :)

features/steps/utils.py
```diff
@@ -139,6 +141,30 @@ def _clean(loggable):
     return data
 
 
+def traverse(resource, path):
```
erikwiffin: I'm late to the party, and this is already merged, but is there a reason you didn't import `traverse` from argonaut.py?

jmandel: Thanks for the review @erikwiffin! This was a net move of the `traverse` code, from Argonaut to utils. I think the idea is it's useful for "more than just Argonaut".

erikwiffin: Ah, thanks! That makes sense.

PR #116, opened 2017-11-09
STU3 Support

mtmcduffie:


config/SMART_STU2.yml
```diff
@@ -0,0 +1,37 @@
...
+        Array.from(document.querySelectorAll('input.glyphbox')).forEach(function(el) { el.disabled=false; });
+        Array.from(document.querySelectorAll('input.glyphbox')).forEach(function(el) { el.checked=true; });
+        Array.from(document.querySelectorAll('button[data-slide=next]')).forEach(function(el) { el.disabled=false; });
+        Array.from(document.querySelectorAll('button[data-slide=next]'))[1].click();
+    - element: "#authorize"
+      optional: True
+      click: True
+      wait: 2
+
+version:
```
erikwiffin: Two thoughts on this config item:

1. I think "version" should be under the "api" namespace.
2. Is there any situation where a single API would be able to handle both DSTU2 and STU3? My point being, do we need version.tags, or would a simple `version: DSTU2` be sufficient?

features/steps/s4s.py
```diff
@@ -120,15 +123,21 @@ def step_impl(context, mu_ccds_query):
...
+
     resource = context.response.json()
 
     assert "resourceType" in resource, \
-           "Resource has no resourceType: {res}".format(res=resource)
+        "Resource has no resourceType: {res}".format(res=resource)
+
+    server_config_key = "API_SERVER_{server_version}".format(server_version=version_name)
+
+    assert server_config_key in current_app.config, \
```
erikwiffin: I'm pretty sure we don't want to pull from current_app here. It probably hasn't been done in a long time, but we (hopefully still) can actually run the entire test suite from the command line just by calling `behave`, and flask wont be loaded at all. If you need to read from ENV variables, I'd suggest adding them to `testsuite/config_reader.py` and having that update the context config.

testsuite/db/.gitignore
```diff
@@ -1,4 +0,0 @@
-# Ignore everything
```
erikwiffin: Do we not need the db directory anymore?

features/steps/s4s.py
mtmcduffie: Ah, good point. I was actually running it from the command line (and adding --tags to test) before I made this change. Will update.

assets/templates/scenarioWasSkipped.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function (options) {
+  // If the first step of the scenario doesn't have a result, it means the scenario was skipped.
+  var skipped = (typeof this.steps[0].result == 'undefined');
```
jmandel: Prefer `===` for string comparisons.

config/SMART_STU2.yml
jmandel: I think we want to cover a case where a single config file says "go here for dstu2-style auth + clinical resources, and there for stu3-style imaging resources". So maybe represent with something like:

```yaml
use_cases:
  security: r2
  clinical: r2
  imaging: r3

api:
  versions:
    r2:
      url: https://url-for-r2-server
      // and other config here if needed
    r3:
      url: https://url-for-r3-server
      // and other config here if needed
```

This is hairy, but I think it captures the "essential complexity" that we want. We could then tag specific features with use case tags like `security`, `clinical`, `imaging` or `financial`... The logic above would mean:

* Any feature tagged with `use_case_security` must also be tagged with `r2`, otherwise we skip it
* Any feature tagged with `use_case_clinical` must also be tagged with `r2`, otherwise we skip it
* Any feature tagged with `use_case_imaging` must also be tagged with `r3`, otherwise we skip it
* Any feature tagged with a `use_case_*` not listed here gets skipped (e.g. `use_case_financial`)

features/00-allergies.feature
```diff
@@ -1,35 +1,46 @@
 @allergies-and-intolerances
```
jmandel: e.g. add `@use_case_clinical @allergies-and-intolerances` to include a "use case" tag.

features/environment.py
```diff
@@ -44,10 +47,16 @@ def before_all(context):
...
     override = getattr(context.config, 'override', os.getenv('CONFIG_OVERRIDE', ''))
+
     vendor_config = get_config(vendor, override)
     vendor_config['auth']['aud'] = vendor_config['api']['url']
+
     context.vendor_config = copy.deepcopy(vendor_config)
 
+    # Restrict the tests to a version specified in the config file for this vendor.
+    if vendor_config.get('version', {}).get('tags'):
+        context.config.tags.ands.append([vendor_config['version']['tags']])
```
jmandel: Update this logic if we do the "use_case" thing.

testsuite/db/.gitignore
jmandel: If this directory is absent, I think our db init fails. (But if that's wrong, I'd love to know and let's kill it.)

config/SMART_STU2.yml
erikwiffin: This reminds me of the "target features, not browsers" advice from the dark days of JS browser compatibility. A list of supported features/use cases makes more sense than a list of supported versions.

One thing I'd be worried about is that I'd give a pretty high degree of likelihood to `https://url-for-r2-server` and `https://url-for-r3-server` requiring different auth credentials, if not entirely different login paths. At that point, they'd need to be broken into separate config files anyways. Why not just start there, instead of having multiple versions of an API inside a single config file?

jmandel: Great point @erikwiffin. Let me clarify!

The goal here is to test a complete "Deployment scenario". So if, let's say, Cerner introduce a DSTU3 and STU3 server that were both available, we'd have *two* Cerner config files, like

## Config for DSTU2 Clinical server
```yaml
use_cases:
  security: r2
  clinical: r2

api:
  versions:
    r2:
      url: https://url-for-r2-server
      // and other config here if needed
```


## Config for STU3 Clinical server
```yaml 
use_cases:
  security: r3
  clinical: r3
api:
  versions:
    r3:
      url: https://url-for-r3-server
      // and other config here if needed
```

The only time you'd need to do what I described above (with mixed content in a single file) would be a split scenario as we're doing for imaging, where you'll go to a DSTU2 auth + clinical server, and a STU3 imaging server (presenting the token you got from the DSUT2-focused auth server).

testsuite/db/.gitignore
mtmcduffie: @jmandel is correct.

PR #120, opened 2017-11-27
Adding Use Cases for testing configurations.

mtmcduffie:


features/environment.py
```diff
@@ -12,7 +13,6 @@
 from testsuite.config_reader import get_vendor_config, get_env_config
 from testsuite.oauth import authorize, factory
 
-
 CACHE_TTL = 60 * 60  # 1 hour
 CCDS_TAGS = {
```
jmandel: This is now a bit of a misnomer -- probably better to call these `FHIR_RESOURCE_TAGS`.

testsuite/config_reader.py
```diff
@@ -27,6 +27,19 @@ def get_vendor_config(vendor, override=''):
...
+    # Interrogate configuration here.
+    try:
+        config['security_version'] = config['use_cases']['security']
+    except KeyError as error:
+        raise Exception("Bad Configuration, could not find version"
+                        "for security under 'use_cases'. (%s)" % error)
+
+    # The config file has the auth and api hierarchies under the version key.
+    # For ease of use, repoint the top level auth and api to be the relevant version.
+    config["auth"] = config["auth"]["versions"][config['security_version']]
```
jmandel: This kind of rewriting is a little bit magical -- it breaks the correspondance between what you see/type, and what's visible to the python code. It might be better to just write accessor functions to handle this, or to populate a new property...

mtmcduffie:
@jmandel - Passed in tags don't override the use cases. So if you passed in a tag corresponding to an EHR use case (say Medication Orders) to a system running the Financial use case, Medication Orders woulds till be skipped. 

If you select Medication Orders and run an EHR use case, everything but Medication Orders will be skipped.

features/environment.py
mtmcduffie: Done!

testsuite/config_reader.py
mtmcduffie: Made this explicit in a new property to refer to a "versioned auth" and "versioned api".

PR #121, opened 2017-11-28
sync-for-science/tracking#133 - Updating use_case logic.

mtmcduffie:
A few straggling changes to the use case logic. Also ensuring the financial use case also checks patient resources.

### DIFF
--- a/features/10-sync-for-science.feature
+++ b/features/10-sync-for-science.feature
@@ -1,22 +1,19 @@
-@s4s
+@use.with_use_case=ehr @s4s
 Feature: Implements all the S4S requirements
 
-    @DSTU2 @STU3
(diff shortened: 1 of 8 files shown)


features/10-sync-for-science.feature
```diff
@@ -1,22 +1,19 @@
-@s4s
+@use.with_use_case=ehr @s4s
```
jmandel: Technically we want two more blocks like this: one `@use.with_use_case=financial` and one  `@use.with_use_case=imaging`. Feel free to fix here or just add an issue to address later!

mtmcduffie: Good point. I'll see what we can fit in one feature file.

https://github.com/sync-for-science/tracking/issues/138

PR #122, opened 2017-12-01
Handle large FHIR payloads

erikwiffin:
Fixes https://github.com/sync-for-science/tracking/issues/132.


testsuite/formatter.py
```diff
@@ -31,6 +31,10 @@ def update_status_data(self):
             if 'systems' in element:
                 element['systems'] = element['systems']()
 
+        # If the entire feature has been skipped, populate the skip_reason
+        if current_feature['status'] == 'skipped':
```
jmandel: Is this grabbing data from the *final* element of the current feature `elements` array? This is slightly surprising -- does the array in general contain only 1 element if the feature has been skipped? If so, can the skip reason logic live under the conditional here?

features/steps/utils.py
```diff
@@ -88,6 +88,28 @@ def get_resource(context, resource):
...
+    with requests.get(url, headers=headers, stream=True) as response:
+        # Limit response size
+        # First check the content-length header, if present. This will prevent downloading overly large
+        # responses, assuming a well configured server.
+        content_length = int(response.headers.get('content-length', 0))
+        assert content_length < CONTENT_LENGTH_LIMIT, \
+                ERROR_CONTENT_LENGTH.format(content_length, CONTENT_LENGTH_LIMIT)
+
+        # Now, complete the download, and check the response size.
+        content_length = len(response.content)
```
jmandel: This property access on `response.content` is what triggers the actual synchronous read/download of the entire body? This is what makes me love/hate Python :-)

testsuite/formatter.py
erikwiffin: From what I've seen, if a feature is skipped, the skip_reason is the same and present in all of the elements. At the very least, the last should be more representative than the first, since if a feature is skipped half-way through, the first element would have no reason, and the last would.

That being said, I did it this way because the element variable was conveniently set, not for that justification.

features/steps/utils.py
erikwiffin: It's actually a pretty sane API http://docs.python-requests.org/en/master/user/advanced/#body-content-workflow, all things considered. If you want an asynchronous stream, use the `iter_content` property. If you want to just download the body as normal, use the `content` property. And this all only gets weird when `stream=True`.

jmandel: Sure -- it's just the notion of a property access (rather than a function call) triggering a blocking step that surprises me.

PR #123, opened 2018-01-24
Initial work on "Summary Report"

mtmcduffie:
I'll have to manually make the migrations for the DB when we merge. If you'd prefer we work in isolation on this branch till this is more mature that's fine, or we can start with this as a base and create tasks to expand.

### DIFF
--- a/testsuite/views.py
+++ b/testsuite/views.py
@@ -76,6 +76,14 @@ def authorized():
     exceptions.abort(500)
 
 
+@app.route('/health_summary/')
+def health_summary():
+
+    latest_results = TestRun.query.all()
+    print(latest_results)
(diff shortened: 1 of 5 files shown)


testsuite/views.py
```diff
@@ -76,6 +78,23 @@ def authorized():
...
 
+@app.route('/health_summary/')
+def health_summary():
+
+    latest_results = {}
+    vendors = get_names()
+
+    for vendor in vendors:
+        last_test = TestRun.query.filter_by(vendor=vendor)\
+            .order_by(desc(TestRun.date_ran)).first()
```
erikwiffin: StackOverflow documentation for `TestRun.date_ran.desc()` - https://stackoverflow.com/a/4187279

PR #124, opened 2018-01-29
Update CMS.yaml

aviars:


config/CMS.yml
```diff
@@ -2,8 +2,8 @@
 api:
   versions:
     STU3:
-      url: https://dev.bluebutton.cms.fhirservice.net/protected/bluebutton/fhir/v1/
-      patient: 3979
+      url: https://sandbox.bluebutton.cms.gov/v1/fhir/
+      patient: 20140000008325
```
jmandel: If your access token response is now returning a `patient`, we can get rid of this `patient` line here!

aviars:
@mtmcduffie  ^^

config/CMS.yml
aviars: @jmandel yes it is returning the patient.  It is also accessible via the `userinfo` endpoint.

jmandel:
@aviars I think the `revoke_url` needs to be set to `htt/o/authorized_tokens/` or something similar. Is that right?

I redeployed with these changes and it seems like the overall signin/approval steps may need re-working; they're failing with:

image

jmandel:
I updated the approval steps too (see https://github.com/sync-for-science/test-suite/commit/b9944a31ee72adbfb2cd78d83a6488d1e84bb52d). Now the deployed version is failing when I it tries to get a token -- it gets an authorization code but can't exchange it for a token.

See [https://tests.demo.syncfor.science](https://tests.demo.syncfor.science/#{"vendor":"CMS","tags":["allergies-and-intolerances","coverage","explanation-of-benefit","immunizations","lab-results","medication-administrations","medication-dispensations","medication-requests","medication-orders","medication-statements","patient-documents","patient-demographics","problems","procedures","smoking-status","vital-signs","s4s","smart","ask-authorization","evaluate-request","exchange-code","use-refresh-token","revoke-authorization"],"override":""})

jmandel:
Hrm, worked the 3rd time. Has the CMS sandbox been unstable?

PR #127, opened 2018-02-13
sync-for-science/tracking#150 - Adding Regex to validate references.

mtmcduffie:


jmandel:
(as documentation of where this came from)

PR #128, opened 2018-02-20
sync-for-science/tracking#156 - Updating UI interaction with CMS controls.

mtmcduffie:


config/CMS.yml
```diff
@@ -25,7 +25,8 @@ auth:
         - element: "#id_password"
           send_keys: "cms-demo"
         - element: "input[type='submit']"
-          click: True
+          execute_script: |
+            jQuery('input[type="submit"]').click();
```
jmandel: Do we know what changed to make the basic "click" stop working? Overall, it's cleaner when we can get by without executing arbitrary js, but if we need this, that's A-OK.

mtmcduffie: My hypothesis is that the Selenium window is small so it scrunches the layout and the header ends up over the input button, which doesn't allow you to click it (I think it might to click based on coordinates?). I can't get it to do it in a FF or Chrome window locally so I think it might be related to Selenium browser drivers.

mtmcduffie: *can't get the bad layout to happen locally

jmandel: Interesting, okay. We do pass in the frame buffer size so this should be under our control.

jmandel: https://github.com/sync-for-science/test-suite/blob/98c604a0225599c0a9f1e3d8433d35382ffef3c0/supervisord.conf#L37

mtmcduffie: Ah, increasing resolution does make it better. Updated PR incoming.

PR #129, opened 2018-02-20
Use screen size instead of javascript

mtmcduffie:


jmandel:
HD FTW! (Fascinating that this works.)

PR #130, opened 2018-02-21
sync-for-science/tracking#154 - Adding a new test that should force a…

mtmcduffie:
… 405 error.

### DIFF
--- a/features/20-exchange-access-token.feature
+++ b/features/20-exchange-access-token.feature
@@ -5,7 +5,7 @@ Feature: App exchanges authorization code for access token
         Given OAuth is enabled
         And I am not logged in
         When I ask for authorization
-        And I exchange my authorization code via a POST request
+        And I exchange my authorization code
         Then the response code should be 200
         And the JSON response will contain access_token
         And the JSON response will contain token_type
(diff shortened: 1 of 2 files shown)


features/20-exchange-access-token.feature
```diff
@@ -5,13 +5,20 @@ Feature: App exchanges authorization code for access token
...
         And the JSON response will contain token_type
         And the JSON response will contain scope
         And the JSON response will contain patient
 
+    Scenario: We cannot use a "GET" request for retrieving an access token
+        Given OAuth is enabled
+        And I am not logged in
+        When I ask for authorization
+        And I exchange my authorization code via a GET request
+        Then the response code should be 405
```
jmandel: I might be less prescriptive on this even the 405 is technically correct, the main invariant that we want to enforce is that the result does succeed -- 4xx or 5xx

features/20-exchange-access-token.feature
```diff
@@ -5,13 +5,20 @@ Feature: App exchanges authorization code for access token
         Given OAuth is enabled
         And I am not logged in
         When I ask for authorization
-        And I exchange my authorization code
+        And I exchange my authorization code via a POST request
```
jmandel: I might just default this  value to POST since it is the only correct thing to do and it will avoid polluting the various places in which we issue this as an ordinary step along the way to some unrelated test (rather than a specific negative test)

PR #136, opened 2018-03-13
sync-for-science/tracking#160 - Initial commit that allows for the sk…

mtmcduffie:
…ipping of steps/features based on configurations.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -150,6 +150,16 @@ def before_feature(context, feature):
     if skip_use_case:
         feature.skip("Feature (%s) not in any use case." % feature)
 
+    try:
+        ignored_steps = context.vendor_config["ignored_steps"][feature.location.filename]
+
+        for step in ignored_steps:
+            if step == "all":
+                feature.skip("Feature (%s) requested skip by vendor." % feature)
+
+    except KeyError as e:
+        logging.info("ignored_steps not configured for this feature.")
+
     tags = list(FHIR_RESOURCE_TAGS.intersection(feature.tags))
 
     if len(tags) > 1:
@@ -179,6 +189,22 @@ def before_scenario(context, scenario):
             if matches and matches.groups()[0] != use_case_version:
                 scenario.skip("Scenario's version (%s) not in Use Case (%s)."
                               % (matches.groups()[0], scenario.feature.use_case))
+
+
+def before_step(context, step):
+
+    context.vendor_skip = False
+
+    try:
+        ignored_steps = context.vendor_config["ignored_steps"][step.location.filename]
+
+        for ignored_step in ignored_steps:
+            print(ignored_step)
+            if step.name == ignored_step:
+                context.vendor_skip = True
+                break
+    except KeyError as e:
+        logging.info("ignored_steps not configured for this step.")
 
 
 class Cache(object):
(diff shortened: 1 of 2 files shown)


mtmcduffie:
Configuration looks like so.

```
ignored_steps:
	features/00-problems.feature:
		- Condition.code is bound to http://snomed.info/sct
	features/00-immunizations.feature:
		- all references will resolve
	features/00-medication-orders.feature:
		- all references will resolve
	features/00-procedures.feature:
		- there exists one date or a time period in Procedure.performedDateTime or Procedure.performedPeriod
	features/00-medication-administrations.feature:
		- all
	features/00-patient-documents.feature:
		- all
	features/00-medication-statements.feature:
		- all references will resolve
		- there exists one date or period in MedicationStatement.effectiveDateTime or MedicationStatement.effectivePeriod
```

features/steps/fhir.py
```diff
@@ -116,6 +116,9 @@ def step_impl(context, field_name):
 
 @then('all references will resolve')
 def step_impl(context):
+    if context.vendor_skip:
+        return
+
```
mtmcduffie: These two lines will need to be in each step that might be skippable. Not a fan of putting something in each step but couldn't find a handy way of skipping the step from within the "before_step" method. @erikwiffin - Any ideas?

erikwiffin: From what I can see, behave doesn't provide a way to skip an individual step like that. What about doing something with tags? Add a `healthcorp-skip` tag to anything that HealthCorp wants to skip, and then when the HealthCorp provider is chosen, dynamically add `--tags=-healthcorp-skip` at the start of the run.

mtmcduffie: That's a good idea but I think building that eventual boolean statement of tags that Behave has to evaluate to determine which steps to run was getting out of hand with 2 decision drivers (User input, Use case). That's why I ended up manually implementing the check for tags for our different "use cases".

mtmcduffie:
I'll note, this will potentially change shortly if we are to abstract the decision logic to another class.

features/environment.py
```diff
@@ -150,6 +150,16 @@ def before_feature(context, feature):
...
 
+    try:
+        ignored_steps = context.vendor_config["ignored_steps"][feature.location.filename]
+
+        for step in ignored_steps:
+            if step == "all":
+                feature.skip("Feature (%s) requested skip by vendor." % feature)
+
+    except KeyError:
+        logging.info("ignored_steps not configured for this feature.")
```
jmandel: Is this an error? Why don't we treat "not configured" as "don't want to skip anything"?

features/environment.py
```diff
@@ -181,6 +191,22 @@ def before_scenario(context, scenario):
...
+
+    try:
+        ignored_steps = context.vendor_config["ignored_steps"][step.location.filename]
+
+        for ignored_step in ignored_steps:
+            print(ignored_step)
+            if step.name == ignored_step:
+                context.vendor_skip = True
+                break
+    except KeyError:
```
jmandel: (Same comment here.)

features/environment.py
mtmcduffie: Do you mean not to rely on a try-except, or to have the log message instead say 'don't want to skip anything'

jmandel: (moving to slack...)

PR #138, opened 2018-04-02
sync-for-science/tracking#165 - Adding a class to decide which steps/resources to test

mtmcduffie:


features/steps/deciders.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def should_run_test(self):
+        return not self.__skip_context()
+
+
+class ResourceDecider:
+    """
+    Determines if a resource should be validated based on Argonaut Value Sets.
+    """
+    argo_vital_codes = ["8716-3",
```
mtmcduffie: This should potentially use the bloom filter to find the valueset that we need to validate.

features/steps/deciders.py
```diff
@@ -0,0 +1,64 @@
...
+        self.resource = resource
+
+    def __needs_argo_validation(self):
+
+        validate = False
+        found_validation_toggle = False
+
+        try:
+            for category_coding in self.resource["category"]["coding"]:
+                if category_coding["code"] == "vital-signs":
```
jmandel: This looks good! One gap is if a given vendor has no results with a `vital-signs` category, we should perhaps throw an error. That could be coded as an additional step in our feature file, like "At least one `vital-signs` category result exists". Does that make sense

features/steps/deciders.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, resource):
+        self.resource = resource
+
+    def __needs_argo_validation(self):
+
+        validate = False
+        found_validation_toggle = False
+
+        try:
```
jmandel: The logic here looks right, but it's a bit hard to follow. I wonder if a "guard" approach might be clearer, like the following pseudocode

```
if resource type != Observation: return True
if category != vitals: return True
if code in "our vitals loincs": return True

return False
```

jmandel: Once we need more logic here, it would make sense to call to a resource-type-specific "should validate" functions, rather than packing all of the logic in here. But that's generalization for later, I think...

features/steps/deciders.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, context):
+        """
+        Behave context of currently running step.
+        :param context: A Behave framework context
+        """
+        self.context = context
+
+    def __skip_context(self):
+        return self.context.vendor_skip
```
jmandel: I'm still a bit concerned that for the 95% use case where we're *not* skipping, every step will be logging out this fact at https://github.com/sync-for-science/test-suite/blob/c346fe0427025e0580f84dfd349430beb67cb366/features/environment.py#L206 (with no logged information about which vendor or step, so the logs will just be filled with the same line over and over). Can we clean this up in this PR too?

jmandel: Also it's important to document how `ignored_steps` in our README and possibly in our example SMART config.

PR #140, opened 2018-04-11
sync-for-science/tracking#166 - Vital Signs Unit validation

mtmcduffie:


features/steps/argonaut.py
```diff
@@ -344,3 +347,58 @@ def step_impl(context, field_name, value):
...
                 values=found,
                 value=value)
+
+
+def vital_unit_validation(field_name, resource, system_url):
+
+    path = field_name.split('.')
+    path.pop(0)
+
+    if path[0] == "valueQuantity":
```
jmandel: Consider collapsing these two branches by creating an intermediate list like `values_to_validate`, which is either  the list of components, or a a list with just the top-level value in it.

features/00-vital-signs.feature
```diff
@@ -52,6 +52,7 @@ Feature: Vital signs
         Then there exists one code in Observation.code
         And there exists a fixed Observation.code.coding.system=http://loinc.org
         And Observation.code is bound to http://fhir.org/guides/argonaut/ValueSet/observation-ccdavitalsignresult
-        # Then when using a panel code...
         Then there exists one patient in Observation.subject
         Then there exists one date and time in Observation.effectiveDateTime or Observation.effectivePeriod
+        Then Proper UCUM codes (http://unitsofmeasure.org) are used if Observation.valueQuantity is present.
```
jmandel: It would be good to add unit validation into this PR as well -- unless you're planning to do it separately?

features/steps/argonaut.py
```diff
@@ -344,3 +359,69 @@ def step_impl(context, field_name, value):
...
+    codes_to_validate = utils.traverse(resource, path + ["code"])
+    values_to_validate = utils.traverse(resource, ["component"])
+
+    if not isinstance(systems_to_validate, list):
+        systems_to_validate = [systems_to_validate]
+
+    if not isinstance(codes_to_validate, list):
+        codes_to_validate = [codes_to_validate]
+
+    if not values_to_validate:
```
jmandel: This is redefined pretty far away its initalization; might just avoid the conditional here and definite it initially as the set of (resource + all of its components)?

features/steps/argonaut.py
```diff
@@ -344,3 +359,69 @@ def step_impl(context, field_name, value):
...
+
+    if any(system != system_url for system in systems_to_validate):
+        return {"resource": resource, "status": "Wrong System"}
+
+    if any(not in_value_set({"code": code}, system_url) for code in codes_to_validate):
+        return {"resource": resource, "status": "Invalid Code"}
+
+    for value in values_to_validate:
+
+        current_code = value["code"]["coding"][0]["code"]
```
jmandel: This is going to break in cases where there are multiple codings (which is allowed) and the "expected" one doesn't come first. Can we update to

```
codes_with_unit_requirements = c["code"] for c in value["code"]["coding"] if c["code"] in vitals_code_lookup
```

mtmcduffie: We can add another issue to have this "collect" errors instead of returning after the first. A problem in many places of the test suite.

features/steps/argonaut.py
mtmcduffie: I don't know a prettier pythonic way of conditionally adding those lists if one of them can be null.

jmandel: There's always 

```py
>>> x = None
>>> y = None
>>> (x or []) + (y or [])
[]
```

features/steps/argonaut.py
```diff
@@ -344,3 +359,68 @@ def step_impl(context, field_name, value):
...
+def vital_unit_validation(field_name, resource, system_url):
+
+    path = field_name.split('.')
+    path.pop(0)
+
+    systems_to_validate = utils.traverse(resource, path + ["system"])
+    codes_to_validate = utils.traverse(resource, path + ["code"])
+    resource_components = utils.traverse(resource, ["component"])
+    values_to_validate = [resource] + \
+        (resource_components if resource_components is not None else [])
```
jmandel: `resource_components or []` should work, no?

PR #142, opened 2018-04-26
sync-for-science/tracking#162

mtmcduffie:
Let's talk about the best point in time to pull the filter. This adds an endpoint to force an update to the filter. We can hook into the Flask App startup scripts potentially, or use the built in before_first_request to pull it.

### DIFF
--- a/testsuite/views.py
+++ b/testsuite/views.py
@@ -128,6 +128,7 @@ def update_bloom_filter():
 
     try:
         r = requests.get(current_app.config['BLOOM_FILTER_URL'])
+        r.raise_for_status()
 
         with open(bloom_file, 'wb') as f:
             f.write(r.content)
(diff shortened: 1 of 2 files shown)


mtmcduffie:
@jmandel - Did you ever get a chance to take a look at this?

testsuite/views.py
```diff
@@ -117,6 +120,19 @@ def health_summary():
     return render_template('health_summary.html', vendors=vendors, latest_results=latest_results)
 
 
+@app.route('/update_bloom', methods=['GET'])
+def update_bloom_filter():
+
+    bloom_file = './data/codes.bf'
+
+    r = requests.get(current_app.config['BLOOM_FILTER_URL'])
+    open(bloom_file, 'wb').write(r.content)
```
jmandel: Just an allocation nit: can you use a `with` block to make it clear there's a `close` happening after the write?  (https://stackoverflow.com/a/5362917 for example discusses)

PR #143, opened 2018-05-24
sync-for-science/tracking#175 - Add test to ensure returned patient I…

abjonnes:
…D matches the queried patient ID

### DIFF
--- a/features/00-patient.feature
+++ b/features/00-patient.feature
@@ -37,3 +37,7 @@ Feature: Patient demographics
         And each Patient.name must have a name.family
         And each Patient.name must have a name.given
         Then there exists one administrative gender in Patient.gender
+
+    Scenario: Returned patient ID matches queried patient ID
+        Given I have a Patient demographics response
+        Then the id field will be the queried ID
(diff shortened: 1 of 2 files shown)


mtmcduffie:
I think it looks good. If it ever becomes a problem we could check if the Patient ID we get from the Oauth exchange is the same as the one we expect from our own config file, but things would have to be seriously wrong if there wasn't harmony between those three points.

Approved!

jmandel:
Oh, that's a really good point @mtmcduffie  -- why wouldn't we check all three? In our typical deployment scenario we should be relying entirely on the value from the OAuth exchange...

jmandel:
Thanks! Perhaps this could be a second, separate test like: 

* $OAuth response must include a patient" (i.e. so if we use an override value from config, it still works but complains loudly with a red mark on the scorecard)

abjonnes:
I factored out two common asserts to reduce code duplication.

jmandel:
Nice -- I like the DRYness, and also the readability improvement. Re: checking patient id in OAuth response vs supplied in config, would you prefer to address this in a separate PR? (I think that could make sense, since currently the value is clobbered when we read the OAuth response, so we'd need to stash it somewhere in order to evaluate the comparison...)

abjonnes:
Yeah I think a separate PR for that makes more sense. I'll work on that as well

jmandel:
cool -- I'd say go ahead and merge this one then whenever you're ready!

PR #144, opened 2018-05-31
172 - Add tests for checking URLs in conformance statement

abjonnes:
sync-for-science/tracking#172

Only `authorize` and `token` are required (http://docs.smarthealthit.org/authorization/conformance-statement/). However, if one of these are missing then the entire test suite fails to run anyway, at least for `SMART_EHR_STU3`...

Should we also validate the URLs themselves?

### DIFF
--- a/features/10-sync-for-science.feature
+++ b/features/10-sync-for-science.feature
@@ -56,3 +56,9 @@ Feature: Implements all the S4S requirements
     Scenario: Server implements Patient documents
         Given I have a valid conformance statement
         And this server supports Patient documents
+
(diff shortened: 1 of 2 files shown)


jmandel:
In general we should make sure these two URLs are present at also make sure they are valid (syntactically, I assume you mean?)

abjonnes:
Yeah, syntactically. Do you have a preferred way to do this? In the other repo I just used `urllib.parse.urlparse` and checked that a scheme existed.

jmandel:
That should work just fine. Certainly an improvement over not looking :-)

PR #145, opened 2018-06-01
176 - Add test for requiring patient ID returned in OAuth response

abjonnes:
sync-for-science/tracking#176

Is the S4S section the correct place for this?

### DIFF
--- a/features/10-sync-for-science.feature
+++ b/features/10-sync-for-science.feature
@@ -56,3 +56,7 @@ Feature: Implements all the S4S requirements
     Scenario: Server implements Patient documents
         Given I have a valid conformance statement
         And this server supports Patient documents
+
+    Scenario: Authorization server returns a patient ID
+        Given OAuth is enabled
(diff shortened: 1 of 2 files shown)


features/10-sync-for-science.feature
```diff
@@ -56,3 +56,7 @@ Feature: Implements all the S4S requirements
     Scenario: Server implements Patient documents
         Given I have a valid conformance statement
         And this server supports Patient documents
+
+    Scenario: Authorization server returns a patient ID
+        Given OAuth is enabled
+        Then a patient ID is supplied
```
jmandel: Might write this as "the access token response includes a patient" where the final word is a variable. Makes it easier for us to add checks for `access_token`, `refresh_token`, etc down the line. Will approve here though and leave this up to you (we can always generalize when we find the need.

abjonnes:
Test functionality is duplicated in other scenarios.

PR #150, opened 2018-06-05
177 - Update smoking status to use code instead of social history, but allow for fallback

abjonnes:
sync-for-science/tracking#177

I played around with a few approaches and this is the clearest one. Adds a new scenario to the smoking status feature which shows a failure when the fallback URI (social history) is used, but the rest of the feature can continue.

### DIFF
--- a/features/steps/s4s.py
+++ b/features/steps/s4s.py
@@ -31,7 +31,7 @@ OAuth2 "{0}" endpoint "{1}" is not a valid URI.
 MU_CCDS_MAPPINGS = {
     'Server metadata': 'metadata',
     'Patient demographics': 'Patient/{patientId}',
-    'Smoking status': 'Observation?category=social-history&patient={patientId}',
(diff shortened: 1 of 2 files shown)


features/steps/s4s.py
```diff
@@ -115,20 +116,41 @@ def step_impl(context, mu_ccds_query):
...
-
+def step_impl(context, mu_ccds_query):  # pylint: disable=W0613
     assert context.response is not None, \
         'Missing response.'
 
     context.execute_steps('then the response code should be 200')
 
-    assert query in context.response.request.url, \
-        'Missing {query}'.format(query=query)
+    assert not context.response.history, \
```
abjonnes: I think the original code here would trigger the failure only during redirects where the redirected URL doesn't contain the initially requested path+query string. I needed to change it for the fallback logic so in its place we can just check the request history, which is nonempty if there have been redirects. It's not exactly the same logic - the new way triggers a failure even if the initially requested path+query are in the redirected URI - but should the server ever be redirecting to some other base URI? Is that allowed in the workflow?

Issue #153, opened 2018-06-13
Fix popup links


abjonnes:
Move away from `data:` URLs.

PR #154, opened 2018-06-13
Fix informational links in test suite results

abjonnes:
#153


assets/main.js
```diff
@@ -191,6 +192,12 @@ $(function () {
     $('#tags').find('.label-info').click();
   });
 
+  $('.container').on('click', '.info-window', function (event) {
+    console.log('running');
+    var data = $(event.currentTarget).data('info');
+    window.open().document.write(info_tmpl({message: atob(data)}));
```
jmandel: Oh, that's very clean!

testsuite/templates/individual_report.txt
```diff
@@ -20,11 +20,12 @@ WARNINGS AND FAILURES
         {% for step in element.steps -%}
             {% if step.result is defined -%}
                 {% if step.result.status != "passed" %}
+                    {% set error = step.result.get('error_message', '').split('\n') %}
 Name : {{ step.name }}
 {% if 'warning' in element.tags -%}
-Warning : {{ step.result.error_message|first }}
+Warning : {{ error|first }}
```
jmandel: Can you add a comment indicating why we want only the first line of `error-message`?

Issue #149, opened 2018-06-05
Update documentation


abjonnes:
Documentation needs improvement and inclusion of Bloom filter details.

PR #155, opened 2018-06-13
#149 - Update documentation and other changes

abjonnes:
#149 

- Expands and corrects documentation
- Add CLI option to download/update Bloom filter; download Bloom filter when container starts
- Move redirect URIs to vendor config files

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,33 +1,28 @@
 # S4S Testsuite
 
-Framework for testing S4S API implementations
+Framework for testing S4S API implementations. The interface is a Flask application, and the tests themselves are run as Celery tasks. The tests are written using the behave framework.
 
-## Build and run in docker
+## Build and run in Docker
+
+We use Docker to set up all the required services and run the test server.
 
     git clone https://github.com/sync-for-science/test-suite
     cd test-suite
     docker build -t tests .
     docker run --rm -it \
-      -e BASE_URL=http://tests.dev.syncfor.science:9003 \
-      -p 5000:5000 \
+      -p 9003:5000 \
       tests
+
+`9003` is the default port, but you can configure it to run on other ports - just be sure to add `-e BASE_URL=http://localhost:<port>` to the Docker command as well.
       
-### Develop in docker
+### Develop in Docker
+
+If you want to develop on the app using Docker, mount the repository on the container. Note that the container still has to be restarted to reflect any changes.
 
     docker run --rm -it \
-      -e BASE_URL=http://tests.dev.syncfor.science:9003 \
-      -p 5000:5000 \
+      -p 9003:5000 \
       -v /host/path/to/test-suite:/usr/src/app \
-      tests \
-      /bin/bash
-      
-... and from inside docker, run "behave" to execute tests.
-
-## Installation
-
-This suite was developed for python 3. To install dependencies run `pip install -r requirements.txt`.
-
-To configure your own project, copy `behave.ini.dist` to `behave.ini`, and change the relevant configuration values.
+      tests
 
 ### Building the js app (for development)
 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -1,69 +1,46 @@
 # S4S Testsuite
 
-Framework for testing S4S API implementations
+Framework for testing S4S API implementations. The interface is a [Flask](http://flask.pocoo.org/) application, and the tests themselves are run as [Celery](http://www.celeryproject.org/) tasks. The tests are written using the [behave](http://behave.readthedocs.io/en/stable/index.html) framework.
 
-## Build and run in docker
+## S4S reference stack
+
+The test suite is one component of the [Sync for Science reference stack](https://github.com/sync-for-science/reference-stack-docker), which contains additional services such as DSTU2 and STU3 FHIR servers. The `SMART EHR` servers in the vendor list of the test suite refer to the servers that run with the reference stack, so they will be unavailable if you choose to run the test suite independently. *Additionally*, the FHIR resource validation steps will fail when the test suite is run independently, because those FHIR servers are used to perform the validation.
```
jmandel: Say "independently (i.e., outside of the full reference stack)" the first time, just because a possible mis-reading of "independently" is "on your own, without our help".

README.md
```diff
@@ -1,69 +1,46 @@
...
-      -v /host/path/to/test-suite:/usr/src/app \
-      tests \
-      /bin/bash
+`9003` is the default port, but you can configure it to run on other ports - just be sure to add `-e BASE_URL=http://localhost:<port>` to the Docker command as well.
       
-... and from inside docker, run "behave" to execute tests.
+### Develop in Docker
 
-## Installation
+If you want to develop on the app using Docker, mount the repository on the container. Note that the container still has to be restarted to reflect any changes.
```
jmandel: Shouldn't python code changes be picked up automatically by Flask when `FLASK_DEBUG=1`? (If not, we should mention this in the note, specifically that flask reloading won't work.

README.md
```diff
@@ -1,69 +1,46 @@
...
-Install npm, then `npm install` and `npm run-script build`. If you're editing
+Install [npm](https://www.npmjs.com/), then `npm install` and `npm run-script build`. If you're editing
 JS, you can run `npm run-script watch` to automatically re-build when files
 change.
 
-## Running the test suite
+## Bloom filter
 
-To run the suite, run `behave` with `VENDOR` like `smart` or another value from https://github.com/sync-for-science/test-suite/tree/master/config.
+The test suite uses a [Bloom filter](https://en.wikipedia.org/wiki/Bloom_filter) to validate known codes. The filter is built with the `data/build_bf.py` script, but requires a number of external databases to be downloaded first. As a convenience, we maintain a prebuilt filter, which is downloaded every time the container starts.
```
jmandel: This is a functionality comment rather than a docs comment, but: it would be good to have the test suite look for the file at a known path and download only if absent. That allows people to mount the file from docker, or to build a new docker image that pre-bakes a version of the file in.

testsuite/views.py
```diff
@@ -172,3 +154,29 @@ def initdb():
     ''' Initialize the database.
     '''
     db.create_all()
+
+
+def update_bloom_filter():
+    bloom_file_status = {}
+    bloom_file = './data/codes.bf'
+
+    try:
```
jmandel: See comment above; or more generally let's talk through ways to avoid the requirement that a developer be online to use/develop the test suite. (Could even be as simple as keeping the current code and adding an env variable to skip `update_bloom_filter` on startup.)

README.md
abjonnes: Turns out this does work, but changes to the `features` directory does not trigger a flask reload (which is why I didn't think it was working), since that directory is not part of the flask app. Behave seems to reread the test files every time anyway, so everything should be good for development with `FLASK_DEBUG`.

abjonnes:
Addressed all @jmandel 's points. I changed the Bloom filter logic to download the filter on container start up if it does not exist already, but manually calling the `/update_bloom_filter` endpoint or running `flask get_bloom_filter --force` will overwrite an existing filter.

